PR #15, opened 2019-10-16
INV-53 Inspector view Actions tab

gordocanchola:
Adding Actions tab to inspector view
  Moving inspector to its own directory


djblue:
image

Looks like you might need to add some padding, but otherwise looks good!

gordocanchola:
> image
> 
> Looks like you might need to add some padding, but otherwise looks good!

Added padding

PR #33, opened 2019-10-29
Json rpc

djblue:


src/main/java/org/codice/ddf/catalog/direct/CatalogMethods.java
```diff
@@ -0,0 +1,539 @@
...
+import org.codice.jsonrpc.JsonRpc;
+import org.codice.jsonrpc.MethodSet;
+import org.geotools.filter.SortByImpl;
+import org.geotools.filter.text.cql2.CQLException;
+import org.geotools.filter.text.ecql.ECQL;
+import org.opengis.filter.Filter;
+import org.opengis.filter.sort.SortBy;
+import org.opengis.filter.sort.SortOrder;
+
+public class CatalogMethods implements MethodSet {
```
rzwiefel: no

rzwiefel: How about CatalogFrameworkMethodSet?

src/main/java/org/codice/ddf/catalog/direct/CatalogMethods.java
```diff
@@ -0,0 +1,539 @@
...
+import org.codice.jsonrpc.Error;
+import org.codice.jsonrpc.JsonRpc;
+import org.codice.jsonrpc.MethodSet;
+import org.geotools.filter.SortByImpl;
+import org.geotools.filter.text.cql2.CQLException;
+import org.geotools.filter.text.ecql.ECQL;
+import org.opengis.filter.Filter;
+import org.opengis.filter.sort.SortBy;
+import org.opengis.filter.sort.SortOrder;
+
```
rzwiefel:
```suggestion

/**
 * A class that represents the set of methods that are callable on the
 * <code>CatalogFramework</code> as a map where the key is the 
 * Json Rpc <code>method</code> string (eg, <code>ddf.catalog/create</code>). 
 * the value of the map is a Method that can be called to dispatch the corresponding 
 * action to the <code>CatalogFramework</code>
 */
```

src/main/java/org/codice/ddf/catalog/direct/CatalogMethods.java
```diff
@@ -0,0 +1,539 @@
...
+  public static final String ATTRIBUTES = "attributes";
+
+  //  private final transient SimpleDateFormat dateFormat = new
+  // SimpleDateFormat(ISO_8601_DATE_FORMAT);
+
+  private final Map<String, DocMethod> METHODS;
+
+  {
+    Builder<String, DocMethod> builder = ImmutableMap.builder();
+    builder.put("ddf.catalog/create", new DocMethod(this::create, ""));
```
rzwiefel:
```suggestion
    builder.put("ddf.catalog/create", new DocMethod(this::create, "Takes the specified parameters (metacards) and calls CatalogFramework::create.`params` takes: `metacards(Required, value: List(Object(`metacardType`:string, `attributes`:Object(Required, `id`: String))) "));
```

src/main/java/org/codice/ddf/catalog/direct/CatalogMethods.java
```diff
@@ -0,0 +1,539 @@
...
+    builder.put("ddf.catalog/query", new DocMethod(this::query, ""));
```
rzwiefel:
```suggestion
    builder.put("ddf.catalog/query", new DocMethod(this::query, "Takes the specified parameters and calls CatalogFramework::query. `params` takes: `cql` (TemporarilyRequired, value: String of cql), `sourceIds` (Optional, value: List of strings, Default: ['ddf.distribution']), `isEnterprise` (Optional, value: boolean, default: false), `properties` (Not yet supported), `startIndex` (Optional, value: integer, default: 1), `pageSize` (Optional, value: integer, default: 250), `sortPolicy` (Optional, value: Object(`propertyName`:String, `sortOrder`: String(ASC or DESC))"));

```

src/main/java/org/codice/ddf/catalog/direct/CatalogMethods.java
```diff
@@ -0,0 +1,539 @@
...
+    builder.put("ddf.catalog/update", new DocMethod(this::update, ""));
```
rzwiefel:
```suggestion
    builder.put("ddf.catalog/update", new DocMethod(this::update, "Takes the specified parameters and calls CatalogFramework::query. `params` takes: `metacards(Required, value: List(Object(`metacardType`:string, `attributes`:Object(Required, `id`: String)))"));
```

src/main/java/org/codice/ddf/catalog/direct/CatalogMethods.java
```diff
@@ -0,0 +1,539 @@
...
+    builder.put("ddf.catalog/delete", new DocMethod(this::delete, ""));
```
rzwiefel:
```suggestion
    builder.put("ddf.catalog/delete", new DocMethod(this::delete, "Takes the specified parameters and calls CatalogFramework::query. `params` takes: `ids` (Required, value: List(String))"));
```

src/main/java/org/codice/ddf/catalog/direct/CatalogMethods.java
rzwiefel: I know we want examples in here, but for the interim here's at least some semi-instructive docstrings

src/main/java/org/codice/ddf/catalog/direct/CatalogMethods.java
```diff
@@ -0,0 +1,539 @@
...
+    return builder.build();
+  }
+
+  private ImmutablePair<Metacard, String> map2Metacard(Map metacard) {
+    if (!(metacard.get(ATTRIBUTES) instanceof Map)) {
+      return of(null, "attributes not exist for params");
+    }
+    Map<String, Object> attributes = (Map) metacard.get(ATTRIBUTES);
+
+    Object rawType = metacard.get("metacard_type");
```
rzwiefel:
```suggestion
    Object rawType = metacard.get("metacardType");
```

src/main/java/org/codice/ddf/catalog/direct/CatalogMethods.java
```diff
@@ -0,0 +1,539 @@
...
+    long timeoutMillis = 0;
+
+    boolean isEnterprise = false;
+    if (params.containsKey("isEnterprise")) {
+      isEnterprise = (boolean) params.get("isEnterprise");
+    }
+
+    List<String> sourceIds = new ArrayList<>();
+    if (params.containsKey("sourceIds")) {
+      // TODO (RCZ) - Handle errors
```
rzwiefel:
```suggestion
      if (!(params.get("sourceIds") instanceof List)) {
        return new Error(
            JsonRpc.INVALID_PARAMS,
            "sourceIds was not a List",
            ImmutableMap.of("path", ImmutableList.of("params", "sourceIds")));
      }
```

rzwiefel: 3epini

src/main/java/org/codice/ddf/catalog/direct/CatalogMethods.java
```diff
@@ -0,0 +1,539 @@
...
+        }
+      case DOUBLE:
+        try {
+          return of(new AttributeImpl(name, Double.parseDouble(value.toString())), null);
+        } catch (NumberFormatException e) {
+          return of(
+              null, String.format("Could not convert value for '%s'. \n%s", name, e.toString()));
+        }
+      default:
+        return null;
```
rzwiefel: No null 🙅‍♂️

rzwiefel:
```suggestion
        return of(new AttributeImpl(name, String.valueOf(value))), null);
```

src/main/java/org/codice/jsonrpc/DocMethod.java
```diff
@@ -0,0 +1,19 @@
+package org.codice.jsonrpc;
+
+import java.util.Map;
+
+public class DocMethod implements Method {
+  private final Method method;
```
rzwiefel: I had to go look around for what a `Method` was. Then i found it and realized oh right its just a Function. 

feelsbadman.jpg

This type is not overrused and i'd still say this is more straightforward just leaving it as `Function<Map<String, Object>, Object>`.

I mean this is just a ✏️ probably i dont' really mind if its left as `Method`. It *is* nice and terse. Just meh. If the definition of the alias was more localized id love it way more 😂

src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -0,0 +1,52 @@
...
+    <argument ref="catalogMethods"/>
+  </bean>
+
+  <bean id="jsonRpcHttpServlet"
+    class="org.codice.jsonrpc.JsonRpcHttpServlet">
+    <argument ref="jsonRpc"/>
+
+    <!--    <cm:managed-properties persistent-id="org.codice.ddf.resourcemanagement.usage"-->
+    <!--      update-strategy="container-managed"/>-->
+    <!--    <property name="monitorLocalSources" value="false"/>-->
```
rzwiefel: We can delete this 😌

src/main/java/org/codice/jsonrpc/DocMethod.java
rzwiefel: okay lets leave it and we'll see how it settles 😌

PR #39, opened 2019-11-05
Added ClusterMap

BernardIgiri:
Added ClusterMap with support for clustering results dynamically as the map zooms and selecting results on the map by mouse click.


src/main/webapp/maps/effects.js
```diff
@@ -0,0 +1,29 @@
+import { useState, useEffect, useRef } from 'react'
+import * as _ from 'lodash'
```
djblue: Can you import the single method you require from lodash?

djblue: Can you import the single method you require from lodash?

PR #46, opened 2019-11-07
Adds Graphql server

gjvera:
The graphql server is now an option that will be set by environment variable and will be set to false by default. 

In order to run/use the graphql server you'll need to
- run`node ./src/main/webapp/intrigue-api/server.js`
- change the boolean flag on line 403 of graphql.js to true
- run revelio as normal


.gitignore
```diff
@@ -3,3 +3,4 @@ target
 .idea
 yarn-error.log
 *.iml
+package-lock.json
```
djblue: Can you remove this, we shouldn't be using npm for this project.

djblue:
I think you might need to update your yarn.lock as well

src/main/webapp/intrigue-api/graphql.js
```diff
@@ -38,10 +38,51 @@ const systemProperties = async () => {
   }
 }
 
-const { send } = createTransport({
-  ...window.location,
-  pathname: ROOT,
-})
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = 0
```
Bdthomson: What is this

PR #64, opened 2019-11-14
Migrates graphql resolvers to use the JSONRPC 

gjvera:


src/main/java/org/codice/ddf/catalog/direct/CatalogMethods.java
```diff
@@ -344,14 +364,25 @@ private Object query(Map<String, Object> params) {
       return new Error(
           INTERNAL_ERROR, "An error occured while running your query - " + e.getMessage());
     }
+    return new ImmutableMap.Builder<String, Object>()
```
gjvera: this well get changed back once we migrate the schema over

src/main/webapp/query-status.js
```diff
@@ -28,7 +28,9 @@ const formatStatus = (status = 'Unknown', info = {}) => {
   }
 
   if (status === 'source.success') {
-    const { hits, count, elapsed } = info
```
djblue:
```javascript
const { hits = '', count = '', elapsed =' ' } = info
```

gjvera:
Don't merge yet. `createMetacard` is getting hung up on something.

PR #71, opened 2019-11-19
Add sorts, source, and result form selection to query settings

samuelechu:
image


src/main/webapp/intrigue-api/schema.graphql
```diff
@@ -186,6 +187,7 @@ type SystemProperties {
 
 type Source {
   available: Boolean
+  local: Boolean
```
djblue: You should add some documentation to this.

PR #75, opened 2019-11-21
added metacardType to query results

rzwiefel:
among other things


rzwiefel:
we close and implement in smaller chunks

PR #76, opened 2019-11-22
Adding point radius location input for Lat/Lon, DMS, UTM, and USNG

BernardIgiri:
- Used GeospatialDraw to implement point radius input.
- Added method to convert GeometryJSON to WKT to simplify filter generation.
- Standardized on GeometryJSON as the data format for location input fields.
- Add storybook of point radius input.
- Laid groundwork for implementing Bounding Box, Polygon, and Line in Lat/Lon, DMS, UTM, and USNG

References the `revelio` branch of GeospatialDraw: https://github.com/connexta/geospatialdraw/tree/revelio

Specifically:
https://github.com/connexta/geospatialdraw/tree/revelio/src/main/webapp/coordinate-editor/react-hooks
and
https://github.com/connexta/geospatialdraw/tree/revelio/src/main/webapp/coordinate-editor/geometry


src/main/webapp/location/point/point-lat-lon.tsx
```diff
@@ -0,0 +1,71 @@
...
+  />
+)
+
+const PointLatLon: React.SFC<Props> = ({
+  value: { lat: initLat, lon: initLon },
+  onChange,
+}) => {
+  const [lat, setLat] = React.useState(initLat)
+  const [lon, setLon] = React.useState(initLon)
+  React.useEffect(
```
samuelechu: Is it necessary for this component, and components for the other coordinate systems to have their own state? The onChange method should be able to update some parent, which will then pass the necessary lat/lon props down to this component.

PR #87, opened 2019-11-27
Added keyword and location input

BernardIgiri:
Added keyword and location input using GeospatialDraw.

keyword-input


package.json
```diff
@@ -12,6 +12,7 @@
     "eslint-plugin-graphql": "^3.1.0",
     "eslint-plugin-import": "^2.18.2",
     "eslint-plugin-react-hooks": "^2.1.2",
+    "fetch-mock": "^8.0.0",
```
djblue: All network interaction should happen though the graphql API. If you are using fetch directly, it's bypassing our network abstraction.

src/main/webapp/location/keyword.tsx
```diff
@@ -0,0 +1,172 @@
...
+        }}
+        getOptionLabel={(suggestion: Suggestion) => suggestion.name}
+        options={suggestions}
+        loading={loading}
+        inputValue={input}
+        onChange={(_e, value) => {
+          if (value) {
+            const { id: keywordId, name } = value
+            ;(async () => {
+              const response = await fetch(`./internal/geofeature?id=${id}`)
```
djblue: We should be accessing data via the graphql API

package.json
BernardIgiri: Okay, I will have to figure out graphql on Monday. Thank you.

BernardIgiri:
Alright, the GraphQL changes should be ready now.

src/main/webapp/intrigue-api/graphql.js
```diff
@@ -7,7 +7,7 @@ import { mergeDeepOverwriteLists } from '../utils'
 const { BatchHttpLink } = require('apollo-link-batch-http')
 const { genSchema, toGraphqlName, fromGraphqlName } = require('./gen-schema')
 
-const ROOT = '/search/catalog/internal'
+const ROOT = 'https://localhost:8993/search/catalog/internal'
```
gjvera: why is this change needed?

BernardIgiri: I kept getting `only absolute urls are supported` errors until I did that in my tests of the GraphQL queries in `keyword.tsx` Not sure why but I just reverted this line.

src/main/webapp/location/geolocation.stories.js
```diff
@@ -84,3 +90,103 @@ Object.keys(editors).forEach(key => {
...
+        [-77.3811, 42.7304],
+        [-77.3811, 42.5304],
+      ],
+    ],
+  },
+  properties: {},
+  id: 'Italy Valley Cemetery',
+}
+const extendedKeywordProps = {
+  getSuggestions: () => JSON.parse(JSON.stringify(keywordSuggestions)),
```
djblue: Why are we `JSON.stringify` then `JSON.parse` here?

src/main/webapp/intrigue-api/schema.graphql
```diff
@@ -277,6 +277,18 @@ type FacetResult {
   value: String
 }
 
+type Suggestion {
+  id: String!
```
djblue:
```suggestion
  id: ID!
```

src/main/webapp/location/keyword.tsx
```diff
@@ -0,0 +1,302 @@
...
+    color,
+  } = properties as KeywordGeoProperties
+  const hasSelection = initKeyword && initKeywordId ? true : false
+  const [input, setInput] = React.useState<string>(initKeyword)
+  const [open, setOpen] = React.useState(false)
+  const [suggestions, setSuggestions] = React.useState<Suggestion[]>([])
+  const [{ keyword, keywordId }, setSelectedKeyword] = React.useState<
+    SelectedKeyword
+  >({})
+  React.useEffect(
```
djblue: I think this code could be simplified. I think we have too many effects happening.

src/main/webapp/location/keyword.tsx
```diff
@@ -0,0 +1,302 @@
...
+        query: FeatureQuery,
+        variables: {
+          id,
+        },
+      }),
+      setFeatureQueryState
+    ).then(r => r.geofeature)
+  }
+  const getSuggestions = async (q: string) => {
+    return runQuery(
```
djblue: I think you might be able to use `useQuery` here. You can leverage the skip option when you don't have enough information to run the query.

src/main/webapp/location/bbox.tsx
```diff
@@ -1,23 +1,11 @@
...
-import { Filter, INTERSECTS, ANY_GEO, GEOMETRY } from './filter'
-import { geoToWKT } from './geo-to-wkt'
+import { coordinates as coordinateEditor } from 'geospatialdraw'
 import FormControl from '@material-ui/core/FormControl'
 import FormLabel from '@material-ui/core/FormLabel'
 import Point from './point'
 import Props from './geo-editor'
 import SpacedLinearContainer from '../spaced-linear-container'
 
-export const generateFilter = (geo: geometry.GeometryJSON): Filter => ({
```
djblue: Why is this being removed?

src/main/webapp/location/keyword.tsx
```diff
@@ -0,0 +1,302 @@
...
+  } = coordinateEditor.geoToPolygonProps(value)
+  const {
+    keyword: initKeyword = '',
+    keywordId: initKeywordId,
+    color,
+  } = properties as KeywordGeoProperties
+  const hasSelection = initKeyword && initKeywordId ? true : false
+  const [input, setInput] = React.useState<string>(initKeyword)
+  const [open, setOpen] = React.useState(false)
+  const [suggestions, setSuggestions] = React.useState<Suggestion[]>([])
```
djblue: I think this state is already part of the apollo cache.

src/main/webapp/location/keyword.tsx
BernardIgiri: Unfortunately useQuery cannot be converted to async functions which is how Keyword is bypassing Apollo in storybook.

src/main/webapp/location/bbox.tsx
BernardIgiri: This is no longer needed, filters can be generated directly from GeometryJSON.

src/main/webapp/location/geolocation.stories.js
BernardIgiri: To get a deep clone that isn't going to modify the original value and avoid any chances of side effects in the Storybook.

src/main/webapp/location/keyword.tsx
Bdthomson: Why can't useQuery be used with async functions? Could you use `useLazyQuery` instead?

BernardIgiri: I'm going to refactor the async functions out. I'll be able to use `useLazyQuery` after that.

src/main/webapp/location/geolocation.stories.js
djblue: Why is the suggestion at risk of being modified though?

src/main/webapp/location/keyword.tsx
djblue: I think the keyword component shouldn't be async. It should accept props that can change over time.

src/main/webapp/location/bbox.tsx
djblue: So the components speak geojson directly? :+1:

src/main/webapp/location/location.tsx
```diff
@@ -0,0 +1,105 @@
...
+    shape: shapes.POLYGON,
+  },
+}
+
+const Location: React.SFC<Props> = ({
+  value,
+  onChange,
+  editorProps = {} as EditorPropsMap,
+}) => {
+  const [locationType, setLocationType] = React.useState<LocationType>(
```
samuelechu: If a component is taking in an onChange method, it shouldn't be keeping track of the state its calling the onChange method with. This data should be passed down as props.

src/main/webapp/location/keyword/container.tsx
```diff
@@ -0,0 +1,144 @@
...
+  value,
+  onChange,
+  minimumInputLength = MINIMUM_INPUT_LENGTH,
+  ...rest
+}) => {
+  const {
+    keyword = '',
+    keywordId = '',
+  } = value.properties as KeywordGeoProperties
+  const [selectedId, setSelectedId] = React.useState('')
```
samuelechu: I think you can move this state higher up in the component tree

samuelechu: both input and selectedId

BernardIgiri: There is no higher up component. The keyword container is just a GeoJSON in GeoJSON out interface. Internal details are not exposed this. This matches the interface provided by the other location input fields allowing for polymorphism.

PR #88, opened 2019-12-02
Adding Bundle/Code Splitting Support 

Schachte:
Adds code splitting to optimize load performance when loading different routes in the application. 

Depends on the following Ace PR:
https://github.com/connexta/ace/pull/62

### DIFF
--- a/src/main/graphql-server/helpers/renderer.js
+++ b/src/main/graphql-server/helpers/renderer.js
@@ -15,7 +15,7 @@ module.exports = async (req, res, next) => {
     if (hasPath(path)) {
       const { originalUrl, clientBundle } = req
       const html = await executeSSR(originalUrl, clientBundle)
-      res.send(html)
+      res.end(html)
     } else {
       next()
     }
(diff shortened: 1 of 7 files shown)


src/main/graphql-server/helpers/renderer.js
```diff
@@ -6,16 +6,16 @@ import { getDataFromTree } from '@apollo/react-ssr'
...
 
 module.exports = async (req, res, next) => {
   const path = req.originalUrl.replace(ROOT_PATH, '')
+
   try {
     if (hasPath(path)) {
       const { originalUrl, clientBundle } = req
       const html = await executeSSR(originalUrl, clientBundle)
-      res.end(html)
+      res.send(html)
```
djblue: I think this should still be `res.end(html)`

src/main/webapp/workspaces/workspaces.js
```diff
@@ -24,14 +24,23 @@ import gql from 'graphql-tag'
 import { useQuery } from '@apollo/react-hooks'
 
 import QueryStatus from '../query-status'
+import { LoadingComponent } from '../routes'
```
djblue: This introduces a cycles in our dependency graph.

PR #100, opened 2019-12-09
Preparing Production Bundles With Docker

Schachte:
- Adds docker file
- Adds docker compose
- Updates express server
- Removes Redux
- Adds more config to hit external DDF

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,14 @@
+.git
+.project
+.vscode
+.settings
+.vs_code
+.settings
+.eslintrc*
+tsconfig*
+yarn.lock
+apollo.config.js
+pom.xml
+build/
+node_modules/
+src/
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,16 @@
+version: '3'
+services:
+  frontend-server:
+    build: .
+    environment:
+      - NODE_ENV=production
+      - DDF_LOCATION=https://docker.for.mac.localhost:8993
+      - EXPRESS_PORT=4000
+    command:
+      [
+        'sh',
+        '-c',
+        'NODE_ENV=$${NODE_ENV} DDF_LOCATION=$${DDF_LOCATION} EXPRESS_PORT=$${EXPRESS_PORT}} node bundle.middleware.js',
+      ]
+    ports:
+      - '4000:4000'
--- a/src/main/graphql-server/server.js
+++ b/src/main/graphql-server/server.js
@@ -1,2 +1,16 @@
-const app = require('./middleware')
-app.listen(4000, () => {})
+require('babel-polyfill')
+
+const express = require('express')
+const router = require('./middleware')
+const compression = require('compression')
+const app = express()
+const port = process.env.EXPRESS_PORT || 4000
+
+app.use(compression())
+app.use('/', router)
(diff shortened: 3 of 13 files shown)


src/main/webapp/app.js
```diff
@@ -27,29 +23,27 @@ const Application = ({ Routes, client }) => {
...
 
-render(Routes, createClient())
+render(Routes, createClientApollo())
 
 if (process.env.NODE_ENV !== 'production') {
   module.hot.accept('./routes', () => {
-    render(require('./routes').default, createClient())
+    render(require('./routes').default, createClientApollo())
   })
   module.hot.accept('./store', () => {
```
gjvera: do we need this now that you removed `store`?

.dockerignore
```diff
@@ -0,0 +1,14 @@
+.git
+.project
+.vscode
+.settings
+.vs_code
+.settings
+.eslintrc*
+tsconfig*
+yarn.lock
+apollo.config.js
```
gjvera: Should apollo.config be ignored in the container?

Schachte: yah, don't need to pass anything into the container once the bundles are built.

docker-compose.yml
```diff
@@ -0,0 +1,16 @@
...
+    build: .
+    environment: 
+      - NODE_ENV=production
+      - DDF_HOST_NAME=docker.for.mac.localhost
+      - DDF_PORT=8993
+      - EXPRESS_PORT=4000
+    command: [
+      "sh", 
+      "-c", 
+      "NODE_ENV=$${NODE_ENV} DDF_HOST_NAME=$${DDF_HOST_NAME} DDF_PORT=$${DDF_PORT} node bundle.middleware.js"
```
djblue: I think you are missing the `EXPRESS_PORT` env variable.

src/main/webapp/intrigue-api/fetch.js
```diff
@@ -22,8 +22,8 @@ const cacheBust = urlString => {
   const { query, ...rest } = url.parse(urlString)
   return url.format({
     ...rest,
-    hostname: 'localhost',
-    port: '8993',
+    hostname: process.env.DDF_HOST_NAME || 'localhost',
```
djblue: Nice :+1:

djblue: Although, do we just want to pass a single env variable with the ddf url?

src/main/webapp/intrigue-api/schema-graphql.js
```diff
@@ -0,0 +1,322 @@
+const typeDefinitions = () => {
```
djblue: This doesn't need to be wrapped in a function.

src/main/graphql-server/server.js
```diff
@@ -1,2 +1,15 @@
...
+const compression = require('compression')
+const app = express()
+const port = process.env.EXPRESS_PORT || 4000
+
+app.use(compression())
+app.use('/', router)
+app.use('/search/catalog', express.static('public'))
+
+app.listen(port, () => {
+  console.log(`Server is running on port ${port}`)
```
djblue: I think you need to add an eslint ignore for this line.

PR #103, opened 2019-12-09
GraphQL Updates

djblue:
- Breaking apart GraphQL resolvers into multiple files
- Co-Locate schema with resolvers
- Examples of capturing GraphQL requests and responses
- Use `schema.json` for `.eslintrc.js` and `apollo.config.js`


src/main/graphql-server/README.md
```diff
@@ -0,0 +1,29 @@
+# GraphQL Tools
+
+## Capture
+
```
willwill96: Is a capture a request associated with some snapshot of code? Might be worth a sentence or two to explain

djblue: I wouldn't say a snapshot of code. More a snapshot of behavior. This is useful to ensure the behavior is as it was at the time of capture.

PR #106, opened 2019-12-10
Replaced old location input with new location input.

BernardIgiri:
Replaced old location input with new GeospatialDraw based location input.


willwill96:
Can you change this to be anyGeo instead: https://github.com/connexta/revelio/blob/master/src/main/webapp/query-filters/filter/dummyDefinitions.tsx#L43

src/main/webapp/query-filters/filter/dummyDefinitions.tsx
```diff
@@ -41,7 +41,7 @@ export const sampleMetacardTypes = {
     enums: [],
   },
   location: {
-    type: 'GEOMETRY',
+    type: 'anyGeo',
     enums: [],
   },
```
willwill96: Sorry, I meant instead of location
```
anyGeo: {
    type: 'GEOMETRY',
    enums: [],
}
```

PR #115, opened 2019-12-10
Adds Lists

samuelechu:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "typescript": "^3.6.4"
   },
   "dependencies": {
-    "@apollo/react-hooks": "^3.1.2",
+    "@apollo/react-hooks": "^3.1.3",
     "@apollo/react-ssr": "^3.1.3",
     "@date-io/date-fns": "^1.3.11",
     "@material-ui/core": "^4.6.1",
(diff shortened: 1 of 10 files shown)


djblue:
Do we need to show the query status for lists?

samuelechu:
> Do we need to show the query status for lists?

DDF shows a status for lists, we could remove it if it's unnecessary though.
image

src/main/webapp/intrigue-api/attributes.json
```diff
@@ -417,7 +417,7 @@
   {
     "id": "lists",
     "multivalued": true,
-    "type": "XML"
+    "type": "JSON"
```
djblue:
```suggestion
    "type": "MetacardAttributes"
```

package.json
```diff
@@ -22,7 +22,7 @@
     "typescript": "^3.6.4"
   },
   "dependencies": {
-    "@apollo/react-hooks": "^3.1.2",
+    "@apollo/react-hooks": "^3.1.3",
```
samuelechu: https://github.com/apollographql/react-apollo/issues/3690

samuelechu: https://github.com/connexta/revelio/pull/115/files#diff-33686393fc5dffe7c4e217ac20f689d0R58

PR #140, opened 2019-12-19
INV-35 Histogram

gordocanchola:
Adding Histogram that will display the number of hits for an
  attribute's values
  Adding a story for histogram
  Adding Histogram to the visualizations menu
  Updating sample-multi-result.json with more sample data

### DIFF
--- a/src/main/webapp/histogram.js
+++ b/src/main/webapp/histogram.js
@@ -1,0 +1,253 @@
+import React, { useState } from 'react'
+import { List, Map, Set, fromJS } from 'immutable'
+
+import Autocomplete from '@material-ui/lab/Autocomplete'
+import TextField from '@material-ui/core/TextField'
+import Typography from '@material-ui/core/Typography'
+
+const AttributeSelect = props => {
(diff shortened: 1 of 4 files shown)


src/main/webapp/histogram.js
```diff
@@ -0,0 +1,253 @@
...
+                {attr}
+              </Typography>
+            </div>
+          )
+        })}
+    </div>
+  )
+}
+
+const renderBars = (total, valueMap, columnWidth) => {
```
samuelechu: total is unused I think

src/main/webapp/histogram.js
```diff
@@ -0,0 +1,253 @@
...
+
+  return (
+    <div style={{ display: 'flex', flexDirection: 'column' }}>
+      <AttributeSelect
+        attributes={options}
+        value={attribute}
+        onChange={setAttribute}
+      />
+      {attrMap.isEmpty() ? null : (
+        <Graph
```
samuelechu: could use props.children in the graph component instead of passing down these as props

PR #148, opened 2019-12-20
User Directory and Log

bdeining:
Changes the working directory to the user's home directory and redirects all output to a log file

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,16 @@
 FROM node:alpine
 
-RUN adduser -S revelio
+RUN adduser -S revelio -G root 
 USER revelio
 
 # Root path for file contents within the container
 WORKDIR '/home/revelio'
 
 # Entry point (express server)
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,15 +1,16 @@
...
-COPY target/server/bundle.middleware.js .
+COPY --chown=revelio:root target/server/bundle.middleware.js .
 
 # Static bundle chunks
-COPY target/webapp/* ./target/webapp/
+COPY --chown=revelio:root target/webapp/* ./target/webapp/
 
-CMD ["node", "bundle.middleware.js"]
+# Start server and create log
+CMD exec node bundle.middleware.js > revelio.log 2>&1
```
djblue: Does docker already store the logs for us? I know docker has a [log command](https://docs.docker.com/engine/reference/commandline/logs/) but I don't know how the storage works.

PR #157, opened 2020-01-08
Adding authentication to revelio

gjvera:
Demo
loginDemo


src/main/webapp/main.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import ReactDOM from 'react-dom'
+import { AppContainer } from '@connexta/ace/react-hot-loader'
+import App from './app'
+
+const render = App => {
+  // TODO: Update render to be hydrate to improve performance
+  ReactDOM.render(
```
willwill96: Do we not need this line anymore? https://github.com/connexta/revelio/blob/master/src/main/webapp/app.js#L12

gjvera: we do, good catch. probably something that I accidentally removed during the rebase

PR #160, opened 2020-01-14
Fixing authorization in playground

gjvera:


src/main/graphql-server/middleware.js
```diff
@@ -68,7 +68,7 @@ const server = new ApolloServer({
   resolvers,
   playground: {
     headers: {
-      Authorization,
+      playgroundAuth,
```
djblue: I don't think we need to pass a custom header, we should be fine to use `Authorization`.

gjvera: I wanted to keep it specific to the playground, and make it evident that's the only thing using it.

djblue: I would say that the `Authorization` header is already well understood where as the new header isn't necessarily adding value, just needless specificity.

PR #163, opened 2020-01-14
Fixing auth between revelio and ddf

gjvera:
Fixes the issue where you can't have a tab of ddf and revelio open at the same time.

### DIFF
--- a/src/main/webapp/login/container.js
+++ b/src/main/webapp/login/container.js
@@ -16,9 +16,8 @@ const useLogin = () => {
 }
 
 const setCookie = cookie => {
-  let parsedCookie = cookie.split(';')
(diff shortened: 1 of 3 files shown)


src/main/webapp/login/container.js
```diff
@@ -16,9 +16,8 @@ const useLogin = () => {
 }
 
 const setCookie = cookie => {
-  let parsedCookie = cookie.split(';')
-  parsedCookie = parsedCookie[0].split('=')
-  cookies.set(parsedCookie[0], parsedCookie[1], { path: '/' })
+  const parsedCookie = cookie.split(';')
```
djblue: I think the cookie parsing and encoding should happen on the graphql side not the client side.

PR #172, opened 2020-01-17
Fixed user privileges in docker container to work in AWS

bdeining:


willwill96:
:pencil2: `Fixed user privile`~~d~~`ges `in docker container to work in AWS`

djblue:
What is this doing exactly?

bdeining:
When the files are copied in docker they need to be owned by the user defined in the docker file (revelio) https://github.com/connexta/revelio/pull/172/files#diff-3254677a7917c6c01f55212f86c57fbfR4 otherwise they are owned by root and the revelio user cannot do anything with them

bdeining:
https://docs.docker.com/engine/reference/builder/#copy

PR #176, opened 2020-01-22
Fixed Draw button in QuerySelector popup

BernardIgiri:
Persisted the isDrawing state when the QuerySelector closes and added a React effect to listen for when drawing completes so that the popup can be reopened with the updated GeometryJSON thus restoring the Draw button functionality.

### DIFF
--- a/src/main/webapp/workspaces/query-selector.js
+++ b/src/main/webapp/workspaces/query-selector.js
@@ -1,13 +1,14 @@
 import Button from '@material-ui/core/Button'
 import Popover from '@material-ui/core/Popover'
 import ExpandMoreIcon from '@material-ui/icons/ExpandMore'
-import React from 'react'
+import React, { useState, useEffect, useRef } from 'react'
 import {
   Actions,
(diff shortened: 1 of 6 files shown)


src/main/webapp/workspaces/query-selector.js
```diff
@@ -30,9 +31,26 @@ const useOpenClose = props => {
...
+  const drawAnchorEl = useRef(null)
+  useEffect(
+    () => {
+      if (isDrawing) {
+        setWasDrawing(true)
+        handleClose()
+      }
+      if (!isDrawing && wasDrawing) {
+        handleOpen({
+          currentTarget: drawAnchorEl.current,
```
willwill96: Can you use anchorEl here?

BernardIgiri: Unfortunately that element is null when the user clicks Apply or Cancel on the `DrawingMenu` even using a reference directly to the edit button doesn't work.

PR #184, opened 2020-01-27
Adding update rpc call/update graphql schema

gjvera:


src/main/webapp/intrigue-api/metacards/metacards.js
```diff
@@ -368,45 +370,44 @@ const createMetacard = async (parent, args, context) => {
...
+  const body = {
+    metacards: [
+      {
+        ids: [id],
+        metacardType: args.attributes.metacard_type,
+        attributes,
+      },
+    ],
+  }
+  const res = await catalog.update(body)
```
gjvera: `renameKeys` here https://github.com/connexta/revelio/pull/184/files#diff-fd0e9d005f52564b980b4cf83cbd2fe3R246 
is causing issues when trying to apply update. need to address

schema.json
```diff
@@ -4920,7 +4920,7 @@
                   "defaultValue": null
                 },
                 {
-                  "name": "attrs",
+                  "name": "attributes",
```
gjvera: Renaming to be more in line the the rpc endpoint

src/main/webapp/intrigue-api/metacards/metacards.js
```diff
@@ -294,28 +294,30 @@ const metacardsByTag = async (parent, args, context) => {
...
+                value: id,
+              },
+              {
+                type: 'LIKE',
+                property: 'metacard-tags',
+                value: '%',
+              },
+            ],
+          },
+          settings: args.settings,
```
willwill96: Instead of creating a map of promises here, can we use a single promise and concat the ids in an OR filter so we only make one network request? Something like: 
```
{
    type: 'OR'
    filters: args.ids.map(id=>({type: '=', property: 'id', id})
}
```

src/main/webapp/intrigue-api/metacards/metacards.js
```diff
@@ -368,45 +370,44 @@ const createMetacard = async (parent, args, context) => {
...
+  const { catalog, fromGraphqlName, toGraphqlName } = context
 
-  const attributes = Object.keys(attrs).map(attribute => {
-    const value = attrs[attribute]
-    return {
-      attribute: fromGraphqlName(attribute),
-      values: Array.isArray(value) ? value : [value],
+  Object.keys(oldMetacardAttrs).forEach(attrs => {
+    if (attributes[attrs] === undefined) {
+      attributes[attrs] = oldMetacardAttrs[attrs]
```
willwill96: We shouldn't modify args in place. I used [setIn](https://immutable-js.github.io/immutable-js/docs/#/setIn) to get around this.

gjvera: attributes are the new attributes that need to be updated. oldMetacardAttrs are all the attributes that exist for said metacard. This is just adding all the necessary keys to `attributes`to make it a full metacard. But I'll look into `setIn`

schema.json
djblue: Looks like the title of the PR should include that we updated the GraphQL schema.

djblue: I would check with other consumers to make sure we don't break them.

gjvera: Updated. I have updated every consumer of this mutation already 👍

PR #186, opened 2020-01-29
Remove date-fns in favor of moment

gordocanchola:
Removing date-fns in favor of moment.
  We were using both and date-fns and moment which do similar things.
  Moment was chosen because there wasn't a really a difference is size
  between the two and we're using moment in other repos


src/main/webapp/time-range.js
```diff
@@ -237,31 +238,29 @@ const DatePicker = props => {
   const [state, setState] = useState(value)
 
   return (
-    <MuiPickersUtilsProvider utils={DateFnsUtils}>
```
djblue: Does this break the stories using time-range?

gordocanchola: Yes, it breaks the time-range stories. Oops.

gordocanchola: fixed the stories

PR #190, opened 2020-01-29
Centralizing Configuration 

Schachte:
- Centralizes configuration instead of growing all the usages of `process.env` everywhere
- Refactor fetch to match more of a middleware type pattern

### DIFF
--- a/src/main/configuration.js
+++ b/src/main/configuration.js
@@ -8,6 +8,7 @@ try {
     env = JSON.parse(fs.readFileSync(configurationPath))
   }
 } catch (err) {
+  //eslint-disable-next-line
   console.error('Error reading in the environment configuration', err)
 }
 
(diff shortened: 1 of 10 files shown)


src/main/configuration.js
```diff
@@ -0,0 +1,25 @@
...
+  if (fs.existsSync(configurationPath)) {
+    env = JSON.parse(fs.readFileSync(configurationPath))
+  }
+} catch (err) {
+  //eslint-disable-next-line
+  console.error('Error reading in the environment configuration', err)
+}
+
+const defaultConfig = {
+  LOG_LEVEL: 'info',
```
Schachte: ✏️ Add docs to each of these to give more context to what each does.

src/main/configuration.js
```diff
@@ -0,0 +1,38 @@
...
+}
+
+const defaultConfig = {
+  // Default log level for debugging
+  LOG_LEVEL: 'info',
+
+  // Request tracing in graphql
+  GRAPHQL_CAPTURE: false,
+
+  // The port we are running the express web server on
```
gjvera: ✏️ Some of these comments seem unnecessary

PR #201, opened 2020-02-04
Add page based search

pklinef:
Adds an example of a URL driven set of pages that act as a hybrid application when combined with SSR.  Also fixes the paging parameters by replacing the CQL paging parameters with the JSON RPC paging parameters. 

http://localhost:8080/search/catalog/search
image

http://localhost:8080/search/catalog/search/results?query=jpg&type=text&startIndex=1&pageSize=15
image

http://localhost:8080/search/catalog/search/results?query=jpg&type=image&startIndex=1&pageSize=15
image

http://localhost:8080/search/catalog/search/results/e0e485c32e8b47a0acd1773074b004d5
image

### DIFF
--- a/src/main/webapp/search/details.js
+++ b/src/main/webapp/search/details.js
@@ -1,0 +1,81 @@
+import React from 'react'
+import { useParams } from 'react-router-dom'
+import { getIn } from 'immutable'
+
+import { useQuery } from '@apollo/react-hooks'
+import gql from 'graphql-tag'
+
+import Paper from '@material-ui/core/Paper'
+import Container from '@material-ui/core/Container'
+import Grid from '@material-ui/core/Grid'
+import LinearProgress from '@material-ui/core/LinearProgress'
+import Typography from '@material-ui/core/Typography'
+
+import Thumbnail from '../thumbnail/thumbnail'
+
+const LoadingComponent = () => <LinearProgress />
+
+const searchByID = gql`
+  query SearchByID($ids: [ID]!) {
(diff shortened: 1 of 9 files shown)


src/main/webapp/search/results.js
```diff
@@ -0,0 +1,218 @@
...
+const GalleryThumbnail = ({ thumbnail }) => {
+  return thumbnail !== undefined ? <Thumbnail src={thumbnail} /> : null
+}
+
+const GalleryItem = props => {
+  const { result } = props
+
+  return (
+    <Grid item xs={3}>
+      <DetailsLink id={result.id}>
```
gordocanchola: It's not obvious that I have to click the browser's back button to get from Details View back to results.

pklinef: That is intentional.  This is an example of a hybrid style application that should look and feel like a natural web page.  Similarly Google and Amazon do not have back buttons.  The URL is always up to date too so you can always copy the URL and get back to the same application state.

src/main/webapp/search/details.js
```diff
@@ -0,0 +1,72 @@
...
+
+  if (loading) {
+    return <LoadingComponent />
+  }
+
+  if (error) {
+    return <div>Error</div>
+  }
+
+  const attributes = data.metacardsById[0].attributes[0]
```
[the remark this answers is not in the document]
pklinef: Good call on `getIn`.  Now I can have it set a value if that value is missing and provide a better error response.

PR #203, opened 2020-02-04
Fix WKT sanitization regex

jrnorth:
Enables geometry sanitization for CQL strings containing WKTs with spaces between the type and the opening parenthesis


willwill96:
:question: Should we add this to the linestring regex as well?

jrnorth:
> ❓ Should we add this to the linestring regex as well?

Yes, added it to point and linestring too.

PR #207, opened 2020-02-10
Adding sharing modal

gjvera:
sharing-demo

Todo:
- Hide share icon for users who are not owners
- Communicate to the user in some way that have read only privilege and disable editing of metacards

### DIFF
--- a/src/main/webapp/index-cards/index-cards.js
+++ b/src/main/webapp/index-cards/index-cards.js
@@ -45,8 +45,8 @@ export const DeleteAction = props => {
 }
 
 export const ShareAction = props => {
-  const { form } = props
-  return <SharingModal form={form} />
+  const { id, title } = props
+  return <SharingModal id={id} title={title} />
 }
 
 export const Actions = CardActions
--- a/src/main/webapp/index-cards/index-cards.js
+++ b/src/main/webapp/index-cards/index-cards.js
@@ -4,7 +4,6 @@ import Typography from '@material-ui/core/Typography'
 
 import IconButton from '@material-ui/core/IconButton'
 
-import ShareIcon from '@material-ui/icons/Share'
 import AddCircleOutlineIcon from '@material-ui/icons/AddCircleOutline'
 import EditIcon from '@material-ui/icons/Edit'
 
@@ -17,6 +16,7 @@ import CardActionArea from '@material-ui/core/CardActionArea'
 import moment from 'moment'
 
 import ConfirmDelete from '../confirm-delete'
+import SharingModal from '../sharing/sharing-modal'
 
 const onClick = action => e => {
   e.preventDefault()
@@ -45,12 +45,8 @@ export const DeleteAction = props => {
 }
 
 export const ShareAction = props => {
-  const { onShare } = props
-  return (
-    <IconButton onClick={onClick(onShare)}>
-      <ShareIcon />
-    </IconButton>
-  )
+  const { form } = props
+  return <SharingModal form={form} />
 }
 
 export const Actions = CardActions
--- a/src/main/webapp/index-cards/index-cards.js
+++ b/src/main/webapp/index-cards/index-cards.js
@@ -45,8 +45,8 @@ export const DeleteAction = props => {
 }
 
 export const ShareAction = props => {
-  const { id, title } = props
-  return <SharingModal id={id} title={title} />
+  const { id, title, metacardType } = props
+  return <SharingModal id={id} title={title} metacardType={metacardType} />
 }
 
 export const Actions = CardActions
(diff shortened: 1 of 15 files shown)


src/main/webapp/intrigue-api/metacards/metacards.js
```diff
@@ -112,6 +121,7 @@ const typeDefs = `
 
     metacardsByTag(tag: String!, settings: QuerySettingsInput): QueryResponse
     metacardsById(ids: [ID]!, settings: QuerySettingsInput): [QueryResponse]
+    metacardSharingAttributes(id: ID, settings: QuerySettingsInput) : Accessors
```
willwill96: I'm not sure we need to have a new resolver for this. I would recommend moving this logic to a react hook on the frontend

src/main/webapp/sharing/container.js
```diff
@@ -0,0 +1,71 @@
...
+const fragment = gql`
+  fragment SharingAttributes on MetacardAttributes {
+    security_access_individuals_read
+    security_access_individuals
+    security_access_administrators
+  }
+`
+const saveMetacardSharing = () => {
+  const mutation = gql`
+    mutation saveMetacardSharing($id: ID!, $attrs: MetacardAttributesInput!) {
```
willwill96: kind of along the same lines as the last comment - if we remove the new resolver and use the generic metacard resolver, we can duplicate the same fragments for saving and getting sharing attributes

src/main/webapp/sharing/user-share-panel.js
```diff
@@ -0,0 +1,114 @@
...
+      .filter(user => user.permission === 'write')
+      .map(user => {
+        return user.userName
+      })
+    attributes['security_access_administrators'] = userState
+      .filter(user => user.permission === 'admin')
+      .map(user => {
+        return user.userName
+      })
+    attributes['metacard_type'] = 'attribute-group'
```
gjvera: remove the hard coding, pass in actual metacard type

src/main/webapp/sharing/sharing.js
```diff
@@ -0,0 +1,63 @@
...
+      >
+        <Tab label="Users" icon={<Person />} />
+        <Tab label="Groups" icon={<Group />} />
+      </Tabs>
+      <Divider />
+      <Typography
+        component="div"
+        role="tabpanel"
+        hidden={0 !== tabValue}
+        id={`scrollable-force-tabpanel-${tabValue}`}
```
willwill96: are these necessary?

src/main/webapp/sharing/container.js
gjvera: My only concern with that is the client handling the data being reconstructed so that the sharing component can use it properly. But I'll look into just using the metacard resolver.

src/main/webapp/index-cards/index-cards.js
```diff
@@ -45,12 +45,8 @@ export const DeleteAction = props => {
...
 
 export const ShareAction = props => {
-  const { onShare } = props
-  return (
-    <IconButton onClick={onClick(onShare)}>
-      <ShareIcon />
-    </IconButton>
-  )
+  const { id, title, metacardType } = props
+  return <SharingModal id={id} title={title} metacardType={metacardType} />
```
gjvera: When we save a metacard, the metacardtype is required to be supplied by the front end otherwise it defaults to the default value for metacard type. This causes saving to break. I'm not sure if there's a better way to get the metacard type to the sharing component other than at this level.

src/main/webapp/sharing/sharing-modal.js
```diff
@@ -0,0 +1,52 @@
+import React from 'react'
+import DialogContent from '@material-ui/core/DialogContent'
+import DialogTitle from '@material-ui/core/DialogTitle'
+import Dialog from '@material-ui/core/Dialog'
+import IconButton from '@material-ui/core/IconButton'
+import Container from './container'
+import ShareIcon from '@material-ui/icons/Share'
+
+export default props => {
+  const [sharing, isSharing] = React.useState(false)
```
willwill96: :pencil2: rename to setSharing

src/main/webapp/sharing/sharing.js
```diff
@@ -0,0 +1,175 @@
...
```
willwill96: Can we leave a comment for what this id does?

src/main/webapp/sharing/user-share-panel.js
```diff
@@ -0,0 +1,37 @@
...
+import FormControl from '@material-ui/core/FormControl'
+import Button from '@material-ui/core/Button'
+import UserAddRow from './user-add-row'
+
+export default props => {
+  const { state, handleChange, removeUser, addUser } = props
+
+  return (
+    <FormControl style={{ width: '100%' }}>
+      <div style={{ maxHeight: '400px', overflow: 'auto', paddingTop: '.5em' }}>
```
willwill96: :pencil2: we use `em` in some places and `rem` in others for sizing. it would be nice to be consistent

src/main/webapp/sharing/sharing.js
gjvera: Removed

PR #222, opened 2020-02-13
Make Basic Search a query builder

samuelechu:


src/main/webapp/basic-search.js
```diff
@@ -29,7 +29,7 @@ import { Location } from './location'
 import { useQuery } from '@apollo/react-hooks'
 import LinearProgress from '@material-ui/core/LinearProgress'
 import gql from 'graphql-tag'
-import { getIn } from 'immutable'
+import { get, getIn, remove } from 'immutable'
```
gjvera: Can you move this up to line 19 with the other imports from immutable? So we don't import the same module on two different lines?

samuelechu: good catch, I'll just have vscode clean up the imports

src/main/webapp/search-forms/editor.tsx
```diff
@@ -90,6 +90,9 @@ const queryToSearch = (query: QueryType) => {
     filterTree,
     srcs: srcs || ['ddf.distribution'],
     sorts: (sorts || []).map(sort => {
+      if (typeof sort !== 'string') {
```
willwill96: What does this do?

samuelechu: the sorts that come from basic search are objects that have the form {attribute, direction} already, rather than a string

willwill96: Can we leave a comment for that?

PR #225, opened 2020-02-14
Addresses some warnings in CI build

gjvera:


src/main/webapp/workspaces/workspaces.js
```diff
@@ -27,7 +27,7 @@ import gql from 'graphql-tag'
 import { useQuery, useMutation } from '@apollo/react-hooks'
 
 import QueryStatus from '../query-status'
-import { BasicSearch, populateDefaultQuery } from '../basic-search'
+import { default as BasicSearch, populateDefaultQuery } from '../basic-search'
```
willwill96: :pencil2: 
```suggestion
import BasicSearch, { populateDefaultQuery } from '../basic-search'
```

src/main/webapp/@storybook/use-state.js
```diff
@@ -1,13 +1,11 @@
 import { action } from '@connexta/ace/@storybook/addon-actions'
 import React from 'react'
 
-export const useState = initialState => {
+export default initialState => {
```
willwill96: Could also just remove the named export

src/main/webapp/search-forms/search-forms.stories.tsx
```diff
@@ -9,7 +9,7 @@ const { SelectionProvider } = require('../react-hooks/use-selection-interface')
 
 import QueryBuilder from '../query-builder/query-builder'
 import { QueryType } from '../query-builder/types'
-const { useState } = require('../@storybook/use-state')
+import useState from '../@storybook/use-state'
```
willwill96: need to use require to avoid the typescript error 
`require('../@storybook/use-state').default`

PR #228, opened 2020-02-17
Updates workspaces query editor to use search form editor

samuelechu:
image

### DIFF
--- a/src/main/webapp/workspaces/workspaces.js
+++ b/src/main/webapp/workspaces/workspaces.js
@@ -1,40 +1,33 @@
-import React, { memo, useState } from 'react'
-
-import LinearProgress from '@material-ui/core/LinearProgress'
-import Typography from '@material-ui/core/Typography'
-
-import TextField from '@material-ui/core/TextField'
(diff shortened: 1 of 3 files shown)


src/main/webapp/workspaces/workspaces.js
```diff
@@ -95,32 +88,21 @@ export const Workspace = () => {
...
-            setQuery(query)
-            onClear()
-            onSearch(query)
-          }}
+  const QueryEditor = props => {
+    const [query, setQuery] = React.useState(props.query)
+    return (
+      <Box height="calc(100vh - 128px)">
+        <Editor
+          //TODO add option to switch between different query builders
```
samuelechu: planning to do this after functionality for adding queries is added

willwill96:
Are the two extra scroll bars (one on the bottom, one on the far right) in the final?

gjvera:
Not sure I agree with using the search form editor modal in workspaces. The test search component seems redundant, along with the visualizations (since there are more visualizations beneath the modal). I think this would confuse alot of users.

PR #239, opened 2020-02-24
Remove duplicate graphql query names

Bdthomson:
### What does this pr do?
- Fixes unique graphql query names

In order to prevent this in the future, you have to make sure the names given to graphql queries are unique. You can see the issue in VSCodes output tab, and the autocomplete for graphql inside of `gql` tags will stop working.

image

image

This fix affects you if you have this plugin (which you should)
image (20)

### DIFF
--- a/src/main/webapp/basic-search.js
+++ b/src/main/webapp/basic-search.js
@@ -497,7 +497,7 @@ const validate = (filterMap = Map()) => {
 }
 
 const MATCHTYPE_ATTRIBUTE = gql`
-  query getAttributeSuggestionListBasicSearch {
+  query getBasicSearchMatchType {
     systemProperties {
       basicSearchMatchType
     }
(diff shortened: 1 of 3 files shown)


src/main/webapp/basic-search.js
```diff
@@ -497,7 +497,7 @@ const validate = (filterMap = Map()) => {
 }
 
 const MATCHTYPE_ATTRIBUTE = gql`
-  query getAttributeSuggestionList {
+  query getAttributeSuggestionListBasicSearch {
```
willwill96: getBasicSearchMatchType

src/main/webapp/user-settings/result-form-select.js
```diff
@@ -10,7 +10,7 @@ import LinearProgress from '@material-ui/core/LinearProgress'
 import ErrorMessage from '../error'
 
 const resultForms = gql`
-  query SearchForms {
+  query SearchFormsAttributeGroup {
```
willwill96: ResultForms

src/main/webapp/basic-search.js
Bdthomson:
```suggestion
  query getBasicSearchMatchType {
```

PR #253, opened 2020-02-25
Limits index card interactions based on permissions

gjvera:
image

image

Still some work to be done here, just putting up a pr for initial impressions.


src/main/webapp/index-cards/index-cards.js
```diff
@@ -36,13 +37,71 @@ export const EditAction = props => {
...
+  tooltip: {
+    backgroundColor: theme.palette.common.white,
+    color: 'rgba(0, 0, 0, 0.87)',
+    boxShadow: theme.shadows[1],
+    fontSize: 11,
+    border: '1px solid #D3D3D3',
+  },
+}))(Tooltip)
+
+export const ReadOnly = props => {
```
willwill96: :pencil2: might be nice to pull into it's own file

gjvera:
Gonna put this on hold until a ux decision is reached

src/main/webapp/components/search-forms/container.tsx
```diff
@@ -99,13 +117,20 @@ const useCreate = () => {
         .filter(({ id }: { id: string }) => id !== data.createMetacard.id)
         .concat(data.createMetacard)
 
+      const user = getIn(cache.readQuery({ query }), ['user'], {})
+      const { email, roles } = user
```
willwill96: Does this break default search forms? Not sure if preferences needs to be updated

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -315,7 +360,10 @@ const useCreate = () => {
   return useMutation(mutation, {
     update: (cache, { data }) => {
       const query = workspaces
-      const { metacardsByTag } = cache.readQuery({ query })
+      const { metacardsByTag, user } = cache.readQuery({ query })
+      securityAttributes.forEach(securityAttr => {
+        data.createMetacard[securityAttr] = []
+      })
```
willwill96: This seems like logic that graphql should handle

src/main/webapp/components/search-forms/container.tsx
gjvera: So I didn't include user prefs here, because there's no way to automatically update the cache when those change since there's no id associated with it. There was an error in the console on master that the `preferences` was missing for the cache update for this. I'm gonna look into a way to see if there's a better way to do it.

src/main/webapp/components/result-forms/route.js
```diff
@@ -171,18 +198,35 @@ const resultForms = gql`
   query ResultForms {
     metacardsByTag(tag: "attribute-group") {
       attributes {
+        security_access_individuals_read
+        security_access_individuals
+        security_access_administrators
+        security_access_groups_read
+        security_access_groups
```
willwill96: Any reason not include these on the respective fragments for workspaces, search forms, and result forms?

gjvera: No, no reason I'll update it to use the fragment

src/main/webapp/components/workspaces/workspaces.js
gjvera: I was having issues with the graphql server not returning attributes that were empty. I can try again though

src/main/webapp/components/search-forms/route.tsx
```diff
@@ -124,13 +144,22 @@ type RouteProps = {
   forms: QueryType[]
   userDefaultForm?: string
   refetch?: () => void
+  userAttributes: UserAttributes
```
willwill96: I don't think the route should need to know about user attributes

src/main/webapp/components/search-forms/container.tsx
willwill96: I'm a little confused by this statement `there's no way to automatically update the cache when those change since there's no id associated with it`. Isn't this a reason why we would manually update it here?

Maybe I'm misunderstanding the problem, but it seems like you could pull preferences and add it to your update of the cache. Something like: 
 ```
user: {
    email,
    roles,
    preferences: {
        {...preferences}
        __typename: 'UserPreferences'
    }
}
```

File: src/main/webapp/components/index-cards/index-cards.js [hunk 6]
```diff
@@ -40,21 +37,25 @@ export const DuplicateAction = props => {
...
+  const { onDelete, message, isWritable } = props
+  return isWritable ? (
+    <ConfirmDelete onDelete={onDelete}>{message}</ConfirmDelete>
+  ) : null
 }
 
 export const ShareAction = props => {
-  const { id, title, metacardType } = props
-  return <SharingModal id={id} title={title} metacardType={metacardType} />
+  const { id, title, metacardType, isAdmin } = props
```
willwill96: :pencil2: Feels like this prop should be called `isShareable`

src/main/webapp/components/search-forms/container.tsx
```diff
@@ -83,6 +90,7 @@ const useCreate = () => {
     mutation CreateSearchForm($attrs: MetacardAttributesInput!) {
       createMetacard(attrs: $attrs) {
         ...SearchFormAttributes
+        security_access_administrators
```
willwill96: don't need this line anymore with the fragment

src/main/webapp/components/result-forms/route.js
```diff
@@ -183,6 +219,7 @@ const useCreate = () => {
     mutation CreateResultForms($attrs: MetacardAttributesInput!) {
       createMetacard(attrs: $attrs) {
         ...ResultFormAttributes
+        security_access_administrators
```
willwill96: don't need with fragment

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -417,6 +466,7 @@ export default () => {
   const { refetch, loading, error, data } = useQuery(workspaces)
   const [create] = useCreate()
   const [_delete] = useDelete()
+  const [message, setMessage] = React.useState(null)
```
samuelechu: Make sure to remove the useState in the Workspaces component https://github.com/connexta/revelio/blob/c99a7d24b76621986724c860c2a4480a84d07aae/src/main/webapp/components/workspaces/workspaces.js#L236

src/main/webapp/components/result-forms/route.js
```diff
@@ -51,8 +58,23 @@ const AddItem = props => {
...
+    securityAttributes,
+    canShare
+  )
+  const readOnly = isReadOnly(
+    canWrite,
+    canShare,
+    securityAttributes,
+    user.email,
+    user.roles
+  )
```
willwill96: Have you considered using one method that takes in a metacard and returns all three of the properties? I'm thinking of a function call like `const { canShare, canWrite, readOnly } = getSecurityPropertiesFromMetacard(form)`

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -477,7 +527,10 @@ export default () => {
       workspaces={workspacesSortedByTime}
       onCreate={onCreate}
       onDelete={onDelete}
+      setMessage={setMessage}
+      message={message}
```
samuelechu: This message would be better rendered separate from the Workspaces component. Instead, pass down a ref that a component down the tree will render to using a portal

gjvera: not sure why this is showing up in this pr might have been from a messed up rebase, but this is already merged

src/main/webapp/components/tooltip/index.js
```diff
@@ -0,0 +1 @@
+export { CustomTooltip } from './tooltip'
```
samuelechu: Can you make this a default export instead of a named one?

samuelechu:
:question: Can you pass in a metacard id into each index card and make the gql queries in the index card actions instead?

src/main/webapp/intrigue-api/metacards/metacards.js
```diff
@@ -414,12 +414,24 @@ const createMetacard = async (parent, args, context) => {
...
   const res = await catalog.create(metacardsToCreate)
-  return renameKeys(toGraphqlName, {
+
+  const securityAttributes = [
+    'security_access_individuals_read',
+    'security_access_individuals',
+    'security_access_groups_read',
+    'security_access_groups',
+  ]
+  let createdMetacard = renameKeys(toGraphqlName, {
```
willwill96: I think this can still be const since you are modifying subfields

src/main/webapp/intrigue-api/metacards/metacards.js
```diff
@@ -414,12 +414,24 @@ const createMetacard = async (parent, args, context) => {
...
   }
   const res = await catalog.create(metacardsToCreate)
-  return renameKeys(toGraphqlName, {
+
+  const securityAttributes = [
+    'security_access_individuals_read',
+    'security_access_individuals',
+    'security_access_groups_read',
+    'security_access_groups',
+  ]
```
willwill96: :pencil2: move below the renameKeys call

gjvera:
> ❓ Can you pass in a metacard id into each index card and make the gql queries in the index card actions instead?

That seems like a lot of extra work when we could do it all in one call. If a user has 20 workspaces/search forms/etc, we  would need to make 21 network calls. 1 for the initial get of all the metacards by tag and then 20 for each index card. I think it's best to keep it as is with 1 network call.

src/main/webapp/components/index-cards/read-only.js
```diff
@@ -0,0 +1,53 @@
...
+export const ReadOnly = props => {
+  const { isReadOnly, indexCardType } = props
+  const readOnlyMessage =
+    'This ' +
+    indexCardType +
+    ' is Read Only. This means you cannot change, delete or share this ' +
+    indexCardType +
+    '. Please duplicate this ' +
+    indexCardType +
+    ' or ask the owner for permissions to make changes.'
```
willwill96: :pencil2: could also use a template string here

Issue #244, opened 2020-02-25
Workspaces


gjvera:
Workspaces page:
~~Share Workspace~~
~~Create Workspace ~~
~~Delete Workspace~~
- Duplicate Workspace
- Subscribe
- Open Workspace in new tab
- Cancel All Searches
- Run All Searches
- Filter Workspaces

willwill96:
No additional PRs should be made under this epic until the workspaces route has been split up: https://github.com/connexta/revelio/issues/280

PR #254, opened 2020-02-26
Added Duplicate To Workspaces

bdeining:
Screen Shot 2020-02-25 at 5 04 53 PM
tied to https://github.com/connexta/revelio/issues/244

depends on https://github.com/connexta/ddf-jsonrpc/pull/11


gjvera:
While I agree the copy icon is the correct icon for duplicate, I don't think that it will be clear to all users. Maybe this one will need a tool tip?

bdeining:
Screen Shot 2020-02-26 at 12 01 25 PM
the old ui uses a similar icon, but i agree a tooltip could be useful

bdeining:
Screen Shot 2020-02-26 at 12 49 15 PM

src/main/webapp/components/index-cards/index-cards.js
```diff
@@ -34,6 +36,16 @@ export const EditAction = props => {
     </IconButton>
   )
 }
+export const DuplicateAction = props => {
+  const { onDuplicate } = props
+  return (
+    <IconButton onClick={onClick(onDuplicate)}>
+      <Tooltip title="Duplicate" aria-label="add">
```
willwill96: :pencil2: 
```suggestion
      <Tooltip title="Duplicate" aria-label="duplicate">
```

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -258,6 +260,39 @@ const workspaceAttributes = gql`
...
   }
 `
+const useClone = () => {
+  const cloneWorkspace = gql`
+    mutation CloneWorkspace($id: ID!) {
+      cloneMetacard(id: $id) {
+        ...WorkspaceAttributes
+        id: id
+        owner: metacard_owner
+        modified: metacard_modified
```
willwill96: I think we should add these attributes to the Workspace gql fragment

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -328,6 +363,7 @@ export default () => {
   const { refetch, loading, error, data } = useQuery(workspaces)
   const [create, redirectId] = useCreate()
   const [_delete] = useDelete()
+  const [_clone] = useClone()
```
willwill96: :pencil2: Any specific reason to add the `_` to the variable name? We had to do that with delete because it's a reserved keyword

willwill96:
I think we should add this interaction to the other routes with card views as well (i.e. result forms and search forms)

src/main/webapp/components/workspaces/workspaces.js
bdeining: no reason, will update

src/main/webapp/components/workspaces/workspaces.js
bdeining: we might want to create a separate issue to clean up all the graphql in this class; they are scattered around and we make it so we dont have to set `metacard_type` differently than other attributes

bdeining:
> I think we should add this interaction to the other routes with card views as well (i.e. result forms and search forms)

I don't believe those are options in regular intrigue, but we can add them to their respective issues.

PR #264, opened 2020-02-28
Add Workspace subscriptions

gjvera:
image
image

### DIFF
--- a/src/main/webapp/components/workspaces/workspaces.js
+++ b/src/main/webapp/components/workspaces/workspaces.js
@@ -220,10 +220,12 @@ const unsubscribeMutation = gql`
     unsubscribeFromWorkspace(id: $id)
   }
 `
+
 const Workspaces = props => {
-  const { workspaces, onCreate, onDelete, setMessage, message } = props
+  const { workspaces, onCreate, onDelete } = props
   const [subscribe] = useMutation(subscribeMutation)
   const [unsubscribe] = useMutation(unsubscribeMutation)
+  const [message, setMessage] = React.useState(null)
   return (
     <IndexCards>
       {message ? (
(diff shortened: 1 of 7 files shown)


src/main/webapp/intrigue-api/metacards/metacards.js
```diff
@@ -433,6 +447,8 @@ const resolvers = {
     createMetacard,
     saveMetacard,
     deleteMetacard,
+    subscribeToWorkspace,
+    unsubscribeFromWorkspace,
```
willwill96: I think we should rename these to be a little more generic
```suggestion
    subscribeToMetacard,
    unsubscribeFromMetacard,
```

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -328,6 +358,7 @@ export default () => {
   const { refetch, loading, error, data } = useQuery(workspaces)
   const [create, redirectId] = useCreate()
   const [_delete] = useDelete()
+  const [message, setMessage] = React.useState(null)
```
willwill96: :pencil2: I think this state should be moved down into Workspaces instead

src/main/webapp/intrigue-api/metacards/metacards.js
gjvera: Well this resolver only hits those endpoints. And their sole job is to subscribe and unsubscribe to/from workspaces. Do you mean having one resolver that handles both? Because idk how to generalize it more than that.

willwill96: Sorry, I didn't realize the endpoint was coupled to workspaces

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -231,6 +257,15 @@ const Workspaces = props => {
                   onDelete={() => onDelete(workspace)}
                   message="This will permanently delete the workspace."
                 />
+                {/*pass in subscribe value*/}
```
gjvera: remove

PR #267, opened 2020-03-02
Results table -> Material Table

gordocanchola:
This PR is for using material-table for our results table.
This would replace the custom built results table
  
Material-table provides built-in functionality such as column reordering, sorting, search bar for filtering.

### DIFF
--- a/src/main/webapp/components/results/results.stories.js
+++ b/src/main/webapp/components/results/results.stories.js
@@ -7,13 +7,13 @@ import genResults from '../../sample-data/gen-results'
 const stories = storiesOf('Result', module)
 
 const attributes = [
-  'Title',
-  'Thumbnail',
-  'Description',
-  'Created',
-  'Modified',
-  'Checksum',
-  'ID',
+  'title',
+  'thumbnail',
+  'description',
+  'created',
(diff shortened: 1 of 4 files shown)


willwill96:
Can you post a screenshot of what this looks like?

gordocanchola:
image

@willwill96 
Working on getting the keyboard interaction with the selection interface working like it did on the old table. 
then I'll look into checkbox column size.

src/main/webapp/components/results/results.stories.js
```diff
@@ -7,13 +7,13 @@ import genResults from '../../gen-results'
...
-  'Modified',
-  'Checksum',
-  'ID',
+  'title',
+  'thumbnail',
+  'description',
+  'created',
+  'modified',
+  'checksum',
+  //  'id',
```
AzGoalie: Why is id commented out?

gordocanchola: I commented that out for testing. Wasn't supposed to commit that.
Good catch. Thank you

gordocanchola: fixed

PR #270, opened 2020-03-03
Extract header section from query builder

samuelechu:
This pr lets us add a three dot menu in the top section of the query editor, which will be used in switching between different search forms.

image

### DIFF
--- a/src/main/webapp/components/query-editor/query-editor.tsx
+++ b/src/main/webapp/components/query-editor/query-editor.tsx
@@ -10,17 +10,25 @@ import React from 'react'
 import useAnchorEl from '../../react-hooks/use-anchor-el'
 import { AttributeDefinition, QueryType } from '../query-builder/types'
 
-const Header = (props: any) => {
-  const [anchorEl, handleOpen, handleClose, open] = useAnchorEl(props)
+type HeaderProps = {
+  queryInteractions?: React.FunctionComponent<{}>
+  query?: QueryType
+  addOptionsRef?: (el: HTMLDivElement) => void
+  onChange: (query: QueryType) => void
+}
+
+const Header = (props: HeaderProps) => {
+  const { query = {} } = props
+  const [anchorEl, handleOpen, handleClose, open] = useAnchorEl()
   return (
     <Box display="flex" style={{ padding: 8 }} alignItems="center">
       <TextField
         fullWidth
-        value={props.query.title}
+        value={query.title}
         variant="outlined"
         label="Search Title"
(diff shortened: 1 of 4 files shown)


src/main/webapp/components/query-editor/query-editor.tsx
```diff
@@ -1,8 +1,45 @@
...
+        fullWidth
+        value={props.query.title}
+        variant="outlined"
+        label="Search Title"
+        onChange={event => {
+          props.onChange(set(props.query, 'title', event.target.value))
+        }}
+        autoFocus
+      />
+      <div style={{ marginLeft: 10 }} ref={props.addOptionsRef} />
```
willwill96: :pencil2: 
```suggestion
      <Box style={{ marginLeft: 10 }} ref={props.addOptionsRef} />
```

src/main/webapp/components/query-editor/query-editor.tsx
```diff
@@ -1,8 +1,45 @@
...
+import IconButton from '@material-ui/core/IconButton'
+import Menu from '@material-ui/core/Menu'
+import TextField from '@material-ui/core/TextField'
+import MoreVertIcon from '@material-ui/icons/MoreVert'
+import { set } from 'immutable'
 import React from 'react'
+import useAnchorEl from '../../react-hooks/use-anchor-el'
 import { AttributeDefinition, QueryType } from '../query-builder/types'
 
+const Header = (props: any) => {
```
willwill96: :pencil2: type out

src/main/webapp/components/query-editor/query-editor.tsx
```diff
@@ -1,8 +1,45 @@
...
+        value={props.query.title}
+        variant="outlined"
+        label="Search Title"
+        onChange={event => {
+          props.onChange(set(props.query, 'title', event.target.value))
+        }}
+        autoFocus
+      />
+      <div style={{ marginLeft: 10 }} ref={props.addOptionsRef} />
+      <IconButton
```
willwill96: Should we hide the icon button if there are no query interactions?

samuelechu: maybe we could do this if we bring this query editor and the search form query editor in sync

src/main/webapp/components/query-editor/query-editor.tsx
samuelechu: Looks like the Box doesn't accept refs :thinking:

PR #272, opened 2020-03-04
Added Text Search form and query creation to workspaces

bdeining:
Adds Text Search form creation to workspace view and implements search and save buttons.  This is a somewhat interim PR in that the inital creation view will be updated to include basic / advanced / custom, but this is a good starting point.  Could use feedback on the form style.
Screen Shot 2020-03-04 at 9 25 10 AM
Screen Shot 2020-03-04 at 9 25 01 AM
Screen Shot 2020-03-04 at 9 24 44 AM
Screen Shot 2020-03-04 at 9 24 39 AM

tied to https://github.com/connexta/revelio/issues/244

### DIFF
--- a/src/main/webapp/components/workspaces/workspaces.js
+++ b/src/main/webapp/components/workspaces/workspaces.js
@@ -5,9 +5,13 @@ import Tab from '@material-ui/core/Tab'
 import Tabs from '@material-ui/core/Tabs'
 import Typography from '@material-ui/core/Typography'
 import gql from 'graphql-tag'
+import Popover from '@material-ui/core/Popover'
+import Button from '@material-ui/core/Button'
 import { getIn } from 'immutable'
 import React, { useState } from 'react'
 import loadable from 'react-loadable'
+import SaveAltIcon from '@material-ui/icons/SaveAlt'
+import TextField from '@material-ui/core/TextField'
 import { Link, Redirect, useParams } from 'react-router-dom'
 import { useQueryExecutor } from '../../react-hooks'
 import {
@@ -23,8 +27,13 @@ import { InlineRetry, SnackbarRetry } from '../network-retry'
 import QueryEditor from '../query-editor'
 import QuerySelector from '../query-selector'
 import QueryStatus from '../query-status'
+import SearchIcon from '@material-ui/icons/Search'
 
 const LoadingComponent = () => <LinearProgress />
+
+const {
+  transformFilterToCQL,
+} = require('../../intrigue-api/metacards/CQLUtils')
 
 let Visualizations = () => null
 if (typeof window !== 'undefined') {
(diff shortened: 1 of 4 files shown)


Bdthomson:
Save/Search buttons should have margin/spacing between them

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -59,6 +68,181 @@ const workspaceById = gql`
...
+        variables: {
+          id: workspaceId,
+          attributes: {
+            queries: [data.createMetacard.id],
+            metacard_type: 'workspace',
+          },
+        },
+        refetchQueries: ['WorkspaceById'],
+      }),
+        setQueryId(data.createMetacard.id)
```
Bdthomson: You probably want to wait to set the query id until after it's been added to the workspace.
```suggestion
    onCompleted: async data => {
      await save({
        variables: {
          id: workspaceId,
          attributes: {
            queries: [data.createMetacard.id],
            metacard_type: 'workspace',
          },
        },
        refetchQueries: ['WorkspaceById'],
      }),
      setQueryId(data.createMetacard.id)
```

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -59,6 +68,181 @@ const workspaceById = gql`
...
+          attrs: {
+            title: title,
+            cql: cql,
+            metacard_tags: ['query'],
+            metacard_type: 'metacard.query',
+          },
+        },
+      })
+    } catch (err) {
+      //eslint-disable-next-line
```
Bdthomson: If I'm not mistaken, console logs get stripped when the build is ran in production mode, so if that's the case you guys should remove the eslint `no-console` rule so you don't have to add these comments everywhere.

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -59,6 +68,181 @@ const workspaceById = gql`
...
+      if (queryId) {
+        // Get query
+        runQuery(queryId)
+      }
+    },
+    [queryId, runQuery]
+  )
+
+  const onClickSearch = async () => {
+    await onCreate(true)
```
Bdthomson: Don't have to wait anymore because nothing is happening afterwards.

```suggestion
  const onClickSearch = () => {
    onCreate(true)
```

src/main/webapp/intrigue-api/metacards/metacards.js
```diff
@@ -123,9 +123,9 @@ const typeDefs = `
     createMetacard(attrs: MetacardAttributesInput!): MetacardAttributes
     saveMetacard(id: ID!, attributes: MetacardAttributesInput!): MetacardAttributes
 
-    # TBD: Should only be used when...
+    # TBD: Should only be used when updating assocations with IDs
```
Bdthomson:
```suggestion
    # Should only be used when updating associations with IDs or other expanded fields (i.e. lists, queries) where you need to set raw values but the schema would otherwise not allow you.
```

src/main/webapp/components/workspaces/workspaces.js
willwill96: I think it's nice to have these explicitly ignored because some people use console logs to debug and we don't have a strict review process, so I think random console logs would get in

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -5,9 +5,13 @@ import Tab from '@material-ui/core/Tab'
 import Tabs from '@material-ui/core/Tabs'
 import Typography from '@material-ui/core/Typography'
 import gql from 'graphql-tag'
```
willwill96: I think we should break the workspaces (index cards) routes and the workspace route into two different files. This one is too big now

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -59,6 +68,181 @@ const workspaceById = gql`
...
+      value: textValue,
+    }
+
+    const cql = transformFilterToCQL(filter)
+    try {
+      await create({
+        variables: {
+          attrs: {
+            title: title,
+            cql: cql,
```
willwill96: I think we should be storing the filterTree and not the cql

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -74,30 +258,58 @@ const Results = ({ results }) =>
...
+      if (data && queryIdToRun) {
+        const queryToRun = data.metacardsById[0].attributes[0].queries.find(
+          q => q.id === queryIdToRun
+        )
+        if (queryToRun !== undefined) {
+          onSearch(queryToRun)
+        }
+      }
+    },
+    [queryIdToRun, data, onSearch]
```
willwill96: does onSearch need to be in this array?

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -59,6 +68,181 @@ const workspaceById = gql`
...
+    onCompleted: data => {
+      save({
+        variables: {
+          id: workspaceId,
+          attributes: {
+            queries: [data.createMetacard.id],
+            metacard_type: 'workspace',
+          },
+        },
+        refetchQueries: ['WorkspaceById'],
```
willwill96: Do we need this?

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -59,6 +68,181 @@ const workspaceById = gql`
...
+  const { runQuery } = React.useContext(WorkspaceContext)
+
+  const onCreate = async () => {
+    const filter = {
+      property: 'anyText',
+      type: 'ILIKE',
+      value: textValue,
+    }
+
+    const cql = transformFilterToCQL(filter)
```
willwill96: We shouldn't have cql anywhere on the frontend. graphql should be responsible for any conversions

src/main/webapp/components/workspaces/workspaces.js
willwill96: Also looks like you don't need the `true` argument because onCreate doesn't take any arguments

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -59,6 +68,181 @@ const workspaceById = gql`
...
+      }
+    }
+    ${workspaceAttributes}
+  `
+
+  const [save] = useMutation(saveMutation)
+
+  const [create] = useMutation(mutation, {
+    onCompleted: data => {
+      save({
```
willwill96: We should be using the update option instead and writing to the apollo cache to avoid the need to add refetchQueries

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -59,6 +68,181 @@ const workspaceById = gql`
...
+  fragment QueryAttributes on MetacardAttributes {
+    title
+    cql
+    metacard_tags
+    metacard_type
+  }
+`
+
+const useCreateQuery = workspaceId => {
+  const [queryId, setQueryId] = React.useState('')
```
willwill96: useCreateQuery shouldn't need to know about the queryId

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -5,9 +5,13 @@ import Tab from '@material-ui/core/Tab'
 import Tabs from '@material-ui/core/Tabs'
 import Typography from '@material-ui/core/Typography'
 import gql from 'graphql-tag'
+import Popover from '@material-ui/core/Popover'
+import Button from '@material-ui/core/Button'
 import { getIn } from 'immutable'
 import React, { useState } from 'react'
 import loadable from 'react-loadable'
+import SaveAltIcon from '@material-ui/icons/SaveAlt'
```
willwill96: Why did you choose to use this icon instead of the ordinary `Save` (floppy disc) icon? Maybe it's just me, but this icon makes me think that something is going to be downloaded, not saved when I see it

src/main/webapp/components/workspaces/workspaces.js
bdeining: Agreed, though it might be best as a separate issue to manage the scope of these changes.

src/main/webapp/components/workspaces/workspaces.js
bdeining: Not sure I follow.  The query ID is needed when the workspace is updated on line 105.  It's also used to execute the useEffect in the case that the user hit `Search` after creating the query itself.

willwill96: useCreateQuery having access to the queryId feels like a code smell to me. I think the logic of associating a query to a workspace should be handled by the graphql side

src/main/webapp/components/workspaces/workspaces.js
willwill96: That's fair. I think it would be a good idea to prioritize this before adding more functionality to workspaces going forward

src/main/webapp/components/workspaces/workspaces.js
Bdthomson: `refetchQueries` is a nice to thing have when you can use it. Normally in our system we can't because our data is stored in SOLR and with soft commit times, a refetchQueries would normally just return old data. However, when you query by id, solr is guaranteed to give you up to date data. In this specific case, the refetchQueries he's running is on a `workspaceById` query which means it will be fine.

I am of the opinion that if you don't have to write this goofy/complicated cache updating logic, you shouldn't have to.

PR #277, opened 2020-03-06
Adds switching between basic and advanced search in query editor

samuelechu:
image

### DIFF
--- a/src/main/webapp/components/query-selector/query-selector.js
+++ b/src/main/webapp/components/query-selector/query-selector.js
@@ -1,12 +1,12 @@
 import Box from '@material-ui/core/Box'
 import Button from '@material-ui/core/Button'
 import ListItemText from '@material-ui/core/ListItemText'
-import MenuItem from '@material-ui/core/MenuItem'
 import Popover from '@material-ui/core/Popover'
 import ExpandMoreIcon from '@material-ui/icons/ExpandMore'
 import SearchIcon from '@material-ui/icons/Search'
-import { set } from 'immutable'
+import { getIn, set } from 'immutable'
 import React, { useEffect, useRef, useState } from 'react'
+import useAnchorEl from '../../react-hooks/use-anchor-el'
 import { BasicSearchQueryBuilder } from '../basic-search/basic-search'
 import {
   Actions,
@@ -16,16 +16,16 @@ import {
 } from '../index-cards'
 import AdvancedSearchQueryBuilder from '../query-builder/query-builder'
 const { useDrawInterface } = require('../../react-hooks')
-import useAnchorEl from '../../react-hooks/use-anchor-el'
 
 const defaultSearchForms = {
-  basic: BasicSearchQueryBuilder,
-  advanced: AdvancedSearchQueryBuilder,
-}
-
-const defaultSearchFormLabels = {
-  basic: 'Basic Search',
-  advanced: 'Advanced Search',
+  basic: {
+    label: 'Basic Search',
+    queryBuilder: BasicSearchQueryBuilder,
+  },
+  advanced: {
+    label: 'Advanced Search',
+    queryBuilder: AdvancedSearchQueryBuilder,
+  },
 }
 
 const getBaseQueryFields = query => {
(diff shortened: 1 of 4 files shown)


src/main/webapp/components/query-selector/query-selector.js
```diff
@@ -31,12 +35,38 @@ const useOpenClose = props => {
...
   const [wasDrawing, setWasDrawing] = useState(false)
   const drawAnchorEl = useRef(null)
+  const [queryBuilder, setQueryBuilder] = React.useState(
+    defaultSearchForms.hasOwnProperty(query.type) ? query.type : 'advanced'
+  )
+
+  const defaultSearchFormInteractions = Object.keys(defaultSearchForms).map(
+    key => {
+      return (
+        <MenuItem key={key} onClick={() => setQueryBuilder(key)}>
```
samuelechu: Should we clear the query each time we change query builders? This is what is done in ddf, I'm guessing because the different searches do not have the same available filters

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -48,8 +48,8 @@ const workspaceById = gql`
         queries {
           id
           title
-          cql
```
samuelechu: I'm removing this field because it doesn't seem to be used elsewhere in this file

src/main/webapp/components/query-selector/query-selector.js
```diff
@@ -82,10 +112,12 @@ const QueryCard = props => {
...
         <Box style={{ margin: '10px' }}>
           <QueryEditor
-            query={props.query}
-            queryBuilder={AdvancedSearchQueryBuilder}
-            onSearch={query => onSearch({ ...props.query, ...query })}
-            onChange={query => onChange({ ...props.query, ...query })}
+            query={query}
+            queryBuilder={defaultSearchForms[queryBuilder]}
+            queryInteractions={queryInteractions}
+            //TODO only grab non-query-builder properties from exisiting query
```
willwill96: Are there side effects to not having this done?

samuelechu: I'm not sure, but it seems safer to grab only the necessary properties rather than everything in the existing query (some which might have been removed by the query builder)

src/main/webapp/components/query-selector/query-selector.js
willwill96: I think we should clear the filterTree at least for now, because there are certain filters that won't work in both builders like you said. We would want to at least preserve the id either way though

src/main/webapp/components/query-selector/query-selector.js
```diff
@@ -12,31 +16,55 @@ import {
...
+  }
+
+  const queryInteractions = [
+    ...defaultSearchFormInteractions(key => {
+      if (key !== query.type) {
+        onChange(set(getBaseQueryFields(query), 'type', key))
+      }
+    }),
+  ]
+
```
willwill96: What is this doing?

samuelechu: It resets the search form, but only does so if the search form type the user selects is different from the current one

src/main/webapp/components/query-selector/query-selector.js
```diff
@@ -12,31 +16,59 @@ import {
...
+  advanced: AdvancedSearchQueryBuilder,
+}
 
-  const handleClose = () => {
-    setAnchorEl(null)
-    if (props.onClose) {
-      props.onClose()
-    }
-  }
+const defaultSearchFormLabels = {
```
willwill96: Have you considered combining defaultSearchForms and defaultSearchFormLabels so that all of the needed information is one place? i.e. something like
```
basic: {
    label: 'Basic Search'
    queryBuilder: BasicSearchQueryBuilder
}
```

samuelechu: At first I thought it would be inconvenient to have nested properties, but that doesn't actually seem like the case.

PR #279, opened 2020-03-06
Fix workspace index cards click area

samuelechu:
image
Wrapping the React Router Link component around index cards results in a click area that extends past the border of the index card. This pr removes the link wrapper and redirects using the useHistory hook instead.

### DIFF
--- a/src/main/webapp/components/workspaces/workspaces.js
+++ b/src/main/webapp/components/workspaces/workspaces.js
@@ -8,7 +8,7 @@ import gql from 'graphql-tag'
 import { getIn } from 'immutable'
 import React, { useState } from 'react'
 import loadable from 'react-loadable'
-import { Link, Redirect, useParams } from 'react-router-dom'
+import { useParams, useHistory } from 'react-router-dom'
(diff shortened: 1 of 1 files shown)


src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -241,33 +241,31 @@ const Workspaces = props => {
         const isSubscribed = workspace.userIsSubscribed
         workspace = workspace.attributes
         return (
-          <Link
+          <IndexCardItem
+            {...workspace}
             key={workspace.id}
-            to={`/workspaces/${workspace.id}`}
-            style={{ textDecoration: 'none' }}
+            onClick={() => props.onSelect(workspace.id)}
```
willwill96: Does the `useHistory` hook and `history.push` work here? 
https://reacttraining.com/react-router/web/api/Hooks/usehistory
Would avoid needing the extra prop

samuelechu: oh wow, good idea! I'll try that

PR #281, opened 2020-03-09
Automatically generate schema using http request

Bdthomson:
- Run `yarn graphql:schema` with dev server running and `schema.json` will automatically be updated.

### DIFF
--- a/scripts/generate-schema-json.js
+++ b/scripts/generate-schema-json.js
@@ -1,0 +1,23 @@
+const fetch = require('node-fetch')
+
+const body = JSON.stringify({
+  operationName: 'IntrospectionQuery',
+  variables: {},
+  query:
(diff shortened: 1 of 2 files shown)


willwill96:
Is there a way to make this automatic on file changes or something like that? I personally don't see much benefit of the command over copy-pasting since you still have to know beforehand to run a script like that.

scripts/generate-schema-json.js
```diff
@@ -0,0 +1,23 @@
...
+  fetch('http://localhost:8080/graphql', {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    body,
+  })
+    .then(res => res.json())
+    .then(json => console.log(JSON.stringify(json, null, 2)))
+    .catch(err => console.log('Error: ', err))
```
willwill96: If we log the error, won't it overwrite the schema file with the error message when you run this command?

PR #283, opened 2020-03-10
Adds text search

samuelechu:
image

### DIFF
--- a/src/main/webapp/components/basic-search/basic-search.js
+++ b/src/main/webapp/components/basic-search/basic-search.js
@@ -271,17 +271,17 @@ export const BasicSearchQueryBuilder = props => {
   }
 
   const AddOptions = () =>
-    addOptionsRef ? (
-      ReactDOM.createPortal(
-        <AddButton addFilter={addFilter} />,
-        props.addOptionsRef
-      )
-    ) : (
-      <React.Fragment>
-        <div style={{ marginLeft: '10px' }} />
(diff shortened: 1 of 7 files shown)


yarn.lock
```diff
@@ -11189,14 +11189,6 @@ minipass@^2.2.1, minipass@^2.3.5:
     safe-buffer "^5.1.2"
     yallist "^3.0.0"
 
-minipass@^2.8.6:
```
samuelechu: Changes that were generated after doing `yarn`.

src/main/webapp/components/basic-search/basic-search.js
```diff
@@ -271,17 +271,17 @@ export const BasicSearchQueryBuilder = props => {
...
-        <div style={{ marginLeft: '10px' }} />
-        <AddButton addFilter={addFilter} />
-      </React.Fragment>
-    )
+    addOptionsRef
+      ? ReactDOM.createPortal(
+          <AddButton addFilter={addFilter} />,
+          props.addOptionsRef
+        )
+      : addOptionsRef !== null && (
```
willwill96: Are we differentiating between null and undefined here? If so that's kind of strange to me

samuelechu: yeah, I thought so too. Maybe I should just make a new component that's independent of basic search

samuelechu:
Closing for now, it looks like text search is currently broken in Intrigue

Issue #215, opened 2020-02-12
Update the sources view to issue faceted queries and display them as sources


bdeining:
Update the sources view to issue faceted queries and display them as sources.  The sources view should be expanded to support treating a specific attribute in the metadata (the values received from issuing a faceted query) as a source.

PR #288, opened 2020-03-12
#215 Update the sources view to issue faceted queries and display them as sources 

mazarag2:
Add results from faceted query to local sources in the search/catalog/sources page

### DIFF
--- a/src/main/webapp/components/sources-info-local/sources-info-local.js
+++ b/src/main/webapp/components/sources-info-local/sources-info-local.js
@@ -1,0 +1,56 @@
+import React from 'react'
+
+import { useQuery } from '@apollo/react-hooks'
+import gql from 'graphql-tag'
+
+import StorageIcon from '@material-ui/icons/Storage'
+import OnlineIcon from '@material-ui/icons/CloudDoneOutlined'
+import OfflineIcon from '@material-ui/icons/OfflineBoltOutlined'
+import { SourcesInfo } from '../sources-info/sources-info'
+import { useSourcePollInterval } from '../sources-info/poll-interval'
+
+const sources = gql`
+  query SourcesInfoLocal($attribute: String!) {
+    facet(attribute: $attribute) {
+      value
+      count
+    }
+    sources {
+      isAvailable
+      sourceId
+      local
+    }
+  }
+`
+const FACET_ATTRIBUTE = 'ext.content-store-name'
+
+const transformProp = source => {
+  source.sourceId = source.value
+  delete source.value
+}
+
+export default () => {
+  const pollInterval = useSourcePollInterval(60000)
+
+  const { loading, data = {}, error, refetch } = useQuery(sources, {
+    pollInterval,
+    variables: { attribute: FACET_ATTRIBUTE },
+  })
+  let listOfSources = []
+  if (!loading) {
+    listOfsources = data.sources.map(source => {
+      source.Icon = source.isAvailable ? OnlineIcon : OfflineIcon
+      transformProp(source)
+      return source
+    })
+    const facetSources = data.facet
+    listOfsources = listOfsources.concat(
+      facetSources.map(facetSource => {
+        facetSource.Icon = StorageIcon
+        facetSource.isAvailable = true
+        return facetSource
+      })
+    )
+  }
+  return <SourcesInfo loading={loading} error={error} refetch={refetch} sources={listOfsources} />
+}
--- a/src/main/webapp/components/sources-info-local/sources-info-local.js
+++ b/src/main/webapp/components/sources-info-local/sources-info-local.js
@@ -38,13 +38,13 @@ export default () => {
   })
   let listOfSources = []
   if (!loading) {
-    listOfsources = data.sources.map(source => {
+    listOfSources = data.sources.map(source => {
       source.Icon = source.isAvailable ? OnlineIcon : OfflineIcon
       transformProp(source)
       return source
     })
     const facetSources = data.facet
-    listOfsources = listOfsources.concat(
+    listOfSources = listOfSources.concat(
       facetSources.map(facetSource => {
         facetSource.Icon = StorageIcon
         facetSource.isAvailable = true
@@ -52,5 +52,12 @@ export default () => {
       })
     )
   }
-  return <SourcesInfo loading={loading} error={error} refetch={refetch} sources={listOfsources} />
+  return (
+    <SourcesInfo
+      loading={loading}
+      error={error}
+      refetch={refetch}
+      sources={listOfSources}
+    />
+  )
 }
--- a/src/main/webapp/components/sources-info/sources-info.js
+++ b/src/main/webapp/components/sources-info/sources-info.js
@@ -40,7 +40,7 @@ const sourcesMessage = offlineCount => {
 
 export const SourcesInfo = props => {
   const sources = props.sources === undefined ? [] : props.sources
-    const { error } = props
+  const { error } = props
   const getIcon = source => (source.isAvailable ? OnlineIcon : OfflineIcon)
   const offlineCount = sources.filter(source => !source.isAvailable).length
   if (props.loading) {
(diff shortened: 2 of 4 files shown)


willwill96:
I don't think we should modify the `sources-info` component directly for these changes. I think that most programs will only want true sources to be displayed

src/main/webapp/components/sources-info/sources-info.js
```diff
@@ -44,41 +55,77 @@ const sourcesMessage = offlineCount => {
...
+    }
+    else{  
+      return OfflineIcon
+    }
+  }
+  const offlineCount = Array.from(SourceSet).filter(source => !source.isAvailable).length
+  
+  const { data, loading, error, refetch } = useQuery(facetedQuery, {
+    variables: { attribute: FACET_ATTRIBUTE },
+  })
```
willwill96: We typically put GraphQL requests in the container to keep separation between logic and presentation. i.e. in here for this component: https://github.com/connexta/revelio/blob/master/src/main/webapp/components/sources-info/sources-info.js#L97

src/main/webapp/components/sources-info/sources-info.js
```diff
@@ -44,41 +55,77 @@ const sourcesMessage = offlineCount => {
...
 }
-
 export const SourcesInfo = props => {
   const sources = props.sources === undefined ? [] : props.sources
-  const offlineCount = sources.filter(source => !source.isAvailable).length
-  const getIcon = source => (source.isAvailable ? OnlineIcon : OfflineIcon)
-  if (props.loading) {
+  const SourceSet = new Set(sources);
+  const getIcon = source => {
+    if(source.__typename == 'FacetResult'){
```
willwill96: We probably shouldn't be using `__typename` inside our react components because it bleeds GraphQL implementation details

willwill96: A better alternative might be to append an extra property to sources from the container and based off where the source came from. I would suggest potentially using an `Icon` property where you specify which icon you want to represent that source. That way, each program can define their own `getIcon` logic from within the container

mazarag2:
if we don't want to modify `source-info` directly to include local sources. should the local sources be its own component somewhere else ?  or have a setting to display local sources in the source page 🤔

src/main/webapp/components/sources-info-local/sources-info-local.js
```diff
@@ -0,0 +1,98 @@
...
+import ListItem from '@material-ui/core/ListItem'
+import ListItemIcon from '@material-ui/core/ListItemIcon'
+import ListItemText from '@material-ui/core/ListItemText'
+import Typography from '@material-ui/core/Typography'
+import OnlineIcon from '@material-ui/icons/CloudDoneOutlined'
+import OfflineIcon from '@material-ui/icons/OfflineBoltOutlined'
+import { SnackbarRetry } from '../network-retry'
+import { SourcesInfo } from '../sources-info/sources-info'
+
+const sources = gql`
```
willwill96: You can combine the `sources` and `facetedQuery` gql tags into one which will remove the need for the isSourceReady logic as both queries will share the same loading and error state:
```
const query = gql`
  query SourcesInfoLocal($attribute: String!) {
    facet(attribute:$attribute) {
      value
      count
    }
    sources {
      isAvailable
      sourceId
      local
    }
    
  }
`
```

src/main/webapp/components/sources-info/sources-info.js
```diff
@@ -47,12 +49,15 @@ const sourcesMessage = offlineCount => {
 
 export const SourcesInfo = props => {
   const sources = props.sources === undefined ? [] : props.sources
-  const offlineCount = sources.filter(source => !source.isAvailable).length
-  const getIcon = source => (source.isAvailable ? OnlineIcon : OfflineIcon)
+  const error = props.error === undefined ? '' : props.error
+  const SourceSet = new Set(sources)
```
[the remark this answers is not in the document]
mazarag2: removing was experimenting with something and no longer need it 👍

src/main/webapp/components/sources-info/sources-info.js
```diff
@@ -80,6 +88,21 @@ export const SourcesInfo = props => {
       </Card>
     </div>
   )
+
+  if (error) {
+    return (
+      <div>
+        {SourcesDropDown}
+        <SnackbarRetry
```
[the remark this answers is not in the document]
mazarag2: The error would be from the faceted Source Query . So if the facetedQuery would fail and the the other Source query succeeded we would show the sources along with a SnackBar retry for the facetedQuery. Since I'm going to consolidate it into one query  I will just return the SnackBarRetry here

src/main/webapp/components/sources-info/sources-info.js
```diff
@@ -91,13 +114,14 @@ const useSourcePollInterval = init => {
 
   return data.systemProperties.sourcePollInterval
 }
-
+export const SourceInfo = (props) => {
+  return <SourceInfoLocal />
+}
```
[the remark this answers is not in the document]
mazarag2: if I remove this export we will only be showing source-info  (so not the component with the facet Query sources-info-local) is that what we want ?

src/main/webapp/components/sources-info-local/sources-info-local.js
```diff
@@ -0,0 +1,77 @@
...
+    sources {
+      isAvailable
+      sourceId
+      local
+    }
+  }
+`
+
+const pollInterval = gql`
+  query SourcePollInterval {
```
willwill96: We ran into an issue a little while back from having gql tags with the same query names: https://github.com/connexta/revelio/pull/239

Can we pull this tag and the `useSourcePollInterval` hook into its own file and share it between `sources-info` and `sources-info-local`?

src/main/webapp/components/sources-info/sources-info.js
```diff
@@ -63,14 +64,13 @@ export const SourcesInfo = props => {
           <List>
             {sources.map(source => {
               const Icon = getIcon(source)
-
               return (
-                <ListItem key={source.sourceId}>
+                <ListItem key={source.sourceId || source.value}>
```
willwill96: :pencil2: Can we normalize the props so we only need one of these fields? (i.e. give base sources the `value` property or give facet sources the `sourceId` property before passing them into this component)

src/main/webapp/components/sources-info/sources-info.js
```diff
@@ -47,12 +47,13 @@ const sourcesMessage = offlineCount => {
 
 export const SourcesInfo = props => {
   const sources = props.sources === undefined ? [] : props.sources
-  const offlineCount = sources.filter(source => !source.isAvailable).length
+  const error = props.error === undefined ? '' : props.error
```
willwill96: :pencil2: don't need to modify the error prop
```suggestion
    const { error } = props
```

src/main/webapp/components/sources-info-local/sources-info-local.js
```diff
@@ -0,0 +1,77 @@
...
+}
+
+export default () => {
+  const pollInterval = useSourcePollInterval(60000)
+
+  const { loading, data = {}, error, refetch } = useQuery(sources, {
+    pollInterval,
+    variables: { attribute: FACET_ATTRIBUTE },
+  })
+  let listOfsources = []
```
willwill96: :pencil2: 
```suggestion
  let listOfSources = []
```

src/main/webapp/components/sources-info-local/sources-info-local.js
```diff
@@ -0,0 +1,77 @@
...
+    return (
+      <SourcesInfo
+        loading={loading}
+        sources={sources}
+        error={error}
+        refetch={refetch}
+      />
+    )
+  }
+  return <SourcesInfo loading={loading} sources={listOfsources} />
```
willwill96: :pencil2: don't need the conditional - can just pass the error in directly
```suggestion
  return <SourcesInfo loading={loading} error={error} loading={loading} sources={listOfsources} />
```

PR #289, opened 2020-03-12
Refactor workspaces

samuelechu:


src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -1,237 +1,70 @@
...
-        module => module.default
-      ),
-    loading: LoadingComponent,
-  })
-}
-
-const workspaceById = gql`
-  query WorkspaceById($ids: [ID]!) {
-    metacardsById(ids: $ids) {
+export const workspaces = gql`
```
willwill96: :pencil2: might be nice to pull the gql tags into their own file

src/main/webapp/components/workspace/index.js
```diff
@@ -0,0 +1 @@
+export { default as Workspace } from './workspace'
```
willwill96: :pencil2: We could probably export these components as default now

PR #297, opened 2020-03-25
Added workspace filtering

AzGoalie:
image
image
image
image

### DIFF
--- a/src/main/webapp/components/workspaces/filter-workspaces.js
+++ b/src/main/webapp/components/workspaces/filter-workspaces.js
@@ -1,5 +1,9 @@
 import React, { useState } from 'react'
-import { List, ListItem, ListItemText, Menu, MenuItem } from '@material-ui/core'
+import List from '@material-ui/core/List'
+import ListItem from '@material-ui/core/ListItem'
+import ListItemText from '@material-ui/core/ListItemText'
+import Menu from '@material-ui/core/Menu'
+import MenuItem from '@material-ui/core/MenuItem'
 
 const FilterWorkspaces = ({ onFilter, userAttributes }) => {
   const [anchorEl, setAnchorEl] = React.useState(null)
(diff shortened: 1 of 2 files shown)


src/main/webapp/components/workspaces/filter-workspaces.js
```diff
@@ -0,0 +1,77 @@
+import React, { useState } from 'react'
+import { List, ListItem, ListItemText, Menu, MenuItem } from '@material-ui/core'
```
gjvera: Can you import these individually rather than deconstructing them? We ran into some build performance regressions a while back due to the overhead added from this. https://github.com/connexta/revelio/pull/131

src/main/webapp/components/workspaces/workspaces.js
```diff
@@ -154,13 +155,23 @@ export default () => {
   const workspacesSortedByTime = workspacesWithSubscriptions.sort(
     (a, b) => (a.attributes.modified > b.attributes.modified ? -1 : 1)
   )
+
+  const [filteredWorkspaces, setFilteredWorkspaces] = useState(
+    workspacesSortedByTime
+  )
```
willwill96: By maintaining the state of workspaces here we lose the ability to take advantage of the apollo cache's store of workspaces and inevitably we will run into syncing issues. Instead, I would recommend maintaining a state of a filter function that gets executed directly on the apollo cache store (i.e. the data object)

src/main/webapp/components/workspaces/filter-workspaces.js
```diff
@@ -0,0 +1,77 @@
...
+    switch (index) {
+      case 0:
+        return onFilter(filterByAnyone)
+      case 1:
+        return onFilter(filterByMe)
+      case 2:
+        return onFilter(filterByNotMe)
+      default:
+        return onFilter(filterByAnyone)
+    }
```
willwill96: returns aren't necessary here

PR #298, opened 2020-03-25
Normalize queries on workspaces

samuelechu:
This pr makes workspace queries always have the same form, instead of being possibly transformed into a executable query. Doing this will aid in not having to transform these queries when doing create or save mutations.

### DIFF
--- a/src/main/webapp/components/query-manager/query-manager.stories.tsx
+++ b/src/main/webapp/components/query-manager/query-manager.stories.tsx
@@ -12,7 +12,7 @@ import { button } from '@storybook/addon-knobs'
 
 stories.add('Basic', () => {
   const [queries, setQueries] = useState(sampleQueries)
-  const [query, setQuery] = useState(sampleQueries[0].id)
+  const [query, setQuery] = React.useState(sampleQueries[0].id)
   const [
     navBarLeftRef,
     setNavBarLeftRef,
@@ -33,10 +33,12 @@ stories.add('Basic', () => {
         queries={queries}
         currentQuery={query}
         QueryEditor={QueryEditor}
-        onChange={queries => setQueries(queries)}
-        onSearch={query => {
-          action('onSearch')(query)
-          setQuery(query.id)
+        onChange={queries => {
+          setQueries(queries)
+        }}
+        onSearch={id => {
+          action('onSearch')(id)
+          setQuery(id)
         }}
       />
     </NavigationBarContext.Provider>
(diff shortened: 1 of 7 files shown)


src/main/webapp/components/workspace/workspace.js
```diff
@@ -74,6 +94,23 @@ export default () => {
   const [queries, setQueries] = useState()
   const { results, status, onSearch, onCancel, onClear } = useQueryExecutor()
 
+  const [queuedSearch, setQueuedSearch] = useState('')
+
+  React.useEffect(
```
samuelechu: This logic will hopefully be taken out in a future pr

src/main/webapp/components/query-manager/query-manager.stories.tsx
```diff
@@ -12,7 +12,7 @@ import { button } from '@storybook/addon-knobs'
 
 stories.add('Basic', () => {
   const [queries, setQueries] = useState(sampleQueries)
-  const [query, setQuery] = useState(sampleQueries[0].id)
+  const [query, setQuery] = React.useState(sampleQueries[0].id)
```
willwill96: :pencil2: unnecessary change

src/main/webapp/components/query-manager/query-manager.tsx
```diff
@@ -64,19 +64,25 @@ const QueryCard = (props: QueryCardProps) => {
...
   const hasQueries = queries && queries.length > 0
   const [anchorEl, handleOpen, handleClose, open] = useAnchorEl()
 
   const queryCards = queries.map(query => (
-    <QueryCard {...props} query={query} key={query.id} />
+    <QueryCard
+      {...props}
+      query={query}
+      key={query.id}
+      onSearch={() => onSearch(query.id!)}
```
willwill96: just curious, what does the syntax of `query.id!` do specifically the `!`?

src/main/webapp/components/query-manager/query-manager.stories.tsx
samuelechu: I didn't want this action to trigger when selecting a query https://github.com/connexta/revelio/blob/master/src/main/webapp/@storybook/use-state.js#L8

src/main/webapp/components/query-manager/query-manager.tsx
samuelechu: I think it assures the compiler that the optional won't be undefined. It was complaining before I added the `!` https://stackoverflow.com/a/42274019

PR #300, opened 2020-03-26
Persist queries added to workspace

samuelechu:


samuelechu:
#301

src/main/webapp/components/workspace/hooks/use-create-query.tsx
```diff
@@ -0,0 +1,44 @@
...
+    metacard_tags
+    type
+    sources
+    sorts
+    detail_level
+    id
+  }
+`
+
+export default (onCreate: any) => {
```
willwill96: I like this hook a lot

src/main/webapp/components/workspace/workspace.js
```diff
@@ -94,22 +96,13 @@ export default () => {
...
-        setQueuedSearch('')
-      }
-    },
-    [queries, queuedSearch]
-  )
+  const saveWorkspace = useSaveWorkspace()
+  const createQuery = useCreateQuery(query => {
+    setQueries(merge([query], queries))
+    setCurrentQuery(query.id)
+    saveWorkspace({ queries: merge([query], queries) })
```
willwill96: :pencil2: could also use es6 syntax here
```suggestion
    saveWorkspace({ queries: [query, ...queries] })
```

src/main/webapp/intrigue-api/metacards/metacards.js
```diff
@@ -460,6 +460,11 @@ const saveMetacard = async (parent, args, context) => {
     )
   }
 
+  if (attributes.queries) {
```
willwill96: :thinking: I think we should consider moving this logic to the react container instead. Normally I would agree that the graphql server should be handling the business logic, but transporting the entire query when you only need the id seems  like a big difference. Do you mind testing the difference on a simulated slow 3g network?

src/main/webapp/components/query-manager/add-query.tsx
```diff
@@ -54,10 +52,9 @@ export default (props: AddQueryProps) => {
         query={query}
         QueryEditor={QueryEditor}
         onSearch={() => {
-          onChange(query)
-          onSearch(query.id!)
+          onCreate(query)
           handleClose()
-          setQuery({ id: new Date().toString() })
+          setQuery({})
```
willwill96: :question: What is this for?

samuelechu: We want the add query popover to reset its query state when the query is added

PR #307, opened 2020-03-28
List Editor 

### LINKED ISSUE #246: Lists
- Create List
~~Run List~~
- Delete List
- Edit List

### LINKED ISSUE #306: Metacard type not show in delete modal.
image
When deleting a metacard whether it be search forms, result forms, or workspaces the modal doesn't get the metacard type and the result is show above. This only happens when you use the icons as opposed to the metacard interaction button.

gjvera:
Fixes #246 
Fixes #306 
image
image
image

image


src/main/webapp/components/lists/list-create.js
```diff
@@ -0,0 +1,136 @@
...
+          alignItems: 'left',
+        }}
+      />
+      <RadioButton
+        label="Limit based on filter"
+        buttonText={[{ text: 'Yes', index: 0 }, { text: 'No', index: 1 }]}
+        defaultButton={1}
+        onChange={e => setList({ ...list, useFilter: e })}
+      />
+      {/*TODO Add Filtering capablities*/}
```
gjvera: Gonna need to talk with @samuelechu or @willwill96 about this piece.

gjvera:
Separate pr gonna be up shortly for actually adding results to lists

src/main/webapp/components/input/intrigue-radio-button.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react'
+import FormLabel from '@material-ui/core/FormLabel'
+import ButtonGroup from '@material-ui/core/ButtonGroup'
+import Button from '@material-ui/core/Button'
+
+export const RadioButton = props => {
+  const { label, buttonText, defaultButton, onChange } = props
+  const [selectedButton, setSelectedButton] = React.useState(defaultButton)
```
willwill96: I think this component should be stateless and have the selected button derived from props

I would also recommend structuring the `buttonText` objects as an array of : `{ label, value }` so you don't need to worry about handling indices

src/main/webapp/components/lists/lists.js
```diff
@@ -8,13 +8,11 @@ import ArchiveIcon from '@material-ui/icons/Archive'
...
+import useAnchorEl from '../../react-hooks/use-anchor-el'
 import React from 'react'
-import {
-  Actions,
-  DeleteAction,
-  EditAction,
-  IndexCardItem,
-} from '../index-cards'
+import { ConfirmDeleteAction as DeleteAction } from '../confirm-delete'
+import { Actions, EditAction, IndexCardItem } from '../index-cards'
```
willwill96: :pencil2: could also import DeleteAction from here instead: https://github.com/connexta/revelio/blob/master/src/main/webapp/components/index-cards/index-cards.js#L50

src/main/webapp/components/lists/lists.js
```diff
@@ -97,20 +125,41 @@ const Lists = props => {
           </ListTitle>
         )
 
-        return (
+        return list['list_bookmarks'] == undefined ? (
```
willwill96: :pencil2:
```suggestion
        return list['list_bookmarks'] === undefined ? (
```

gjvera: I've been having alot of issues with list['list_bookmarks'] being null as opposed to undefined.

src/main/webapp/components/lists/list-create.js
```diff
@@ -0,0 +1,153 @@
...
+  const { prevList, onClose, onSave } = props
+  const isNewList = prevList === undefined
+
+  const [list, setList] = React.useState({
+    id: isNewList ? newUuid().replace(/-/g, '') : prevList.id,
+    list_icon: isNewList ? 'folder' : prevList.list_icon,
+    list_bookmarks: isNewList ? [] : prevList.list_bookmarks,
+    query: isNewList ? undefined : prevList.query,
+    title: isNewList ? 'Untitled List' : prevList.title,
+  })
```
willwill96, line 72: This logic is kind of awkward. I think it would read better like this:
```suggestion
  const { onClose, onSave } = props
  const [list, setList] = React.useState(props.list || {
    id: newUuid().replace(/-/g, ''),
    list_icon: 'folder' ,
    list_bookmarks: [],
    query: undefined,
    title: 'Untitled List',
  })
```

src/main/webapp/components/lists/list-create.js
```diff
@@ -0,0 +1,153 @@
...
+          display: 'flex',
+          flexDirection: 'row',
+          width: '100%',
+          justifyContent: 'space-between',
+        }}
+      >
+        <Button
+          variant="outlined"
+          color="primary"
+          style={{ width: '49%' }}
```
willwill96, line 134: `50%`?

src/main/webapp/components/lists/list-create.js
```diff
@@ -0,0 +1,153 @@
...
+          onClick={() => onClose()}
+        >
+          Cancel
+        </Button>
+        <Button
+          variant="contained"
+          color="secondary"
+          style={{ width: '49%' }}
```
willwill96, line 142: `50%`?

src/main/webapp/components/lists/list-create.js
```diff
@@ -0,0 +1,153 @@
...
+          onClick={() => {
+            onSave(list, isNewList)
```
willwill96, line 144: I don't think the editor should need to know whether the list is new or not. I would expect `onSave` prop to handle whether it's being created or updated

src/main/webapp/components/lists/list-create.js
gjvera: I put it at 49% instead of 50% so that there was a little space between the buttons. When it was 50% there was no gap and it looked a little weird to me

src/main/webapp/components/workspace/workspace.js
```diff
@@ -217,6 +220,10 @@ export default () => {
                 }, [])
                 setListResults(results)
               }}
+              onSave={lists => {
```
samuelechu, line 223: I think it would be better if you only saved the lists currently stored in `listState`, and pass down another callback to change list state. You are currently changing the list state and also saving a version not in sync with the one stored on the workspaces component. I think that separating these concerns and keeping a single source of truth will make things easier to reason about as more functionality is added.

src/main/webapp/components/lists/lists.js
```diff
@@ -97,58 +120,103 @@ const Lists = props => {
...
-          transformOrigin={{
-            vertical: 'top',
-            horizontal: 'left',
-          }}
+  return lists ? (
+    <React.Fragment>
+      <div style={{ display: 'flex' }}>
+        {selected
+          ? ListCards.filter(list => list.props.children[0].key === selected)[0]
+          : ListCards[0].props.children}
```
willwill96, line 170: I think it would be nice if we could achieve this logic without using children props. It seems like you could move the key from the IndexCardItem to the React.Fragment on lines 137-138

:question: Why return the ListCards element directly in the first case, but return the props.children in the second case?

src/main/webapp/components/workspace/workspace.js
```diff
@@ -106,13 +108,24 @@ export default () => {
...
   const { loading, error, data } = useQuery(workspaceById, {
     variables: { ids: [id] },
     onCompleted: data => {
       const queries = data.metacardsById[0].attributes[0].queries
+      const lists = data.metacardsById[0].attributes[0].lists
+      lists &&
+        setLists(
+          lists.map(list => {
+            const { __typename, ...rest } = list
+            return list
```
willwill96, line 126: should we return rest here?

src/main/webapp/components/input/intrigue-radio-button.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import FormLabel from '@material-ui/core/FormLabel'
+import ButtonGroup from '@material-ui/core/ButtonGroup'
+import Button from '@material-ui/core/Button'
+
+export const RadioButton = props => {
```
willwill96, line 6: :pencil2: default export seems more appropriate

src/main/webapp/components/workspace/workspace.js
gjvera: Yup good catch.

src/main/webapp/components/lists/lists.js
gjvera: Each ListCard has both the list and the pop over as children since they are both wrapped in a fragment. I'm still returning the ListCard on the first one too. It's just using the key from the list to determine if it's selected.

PR #316, opened 2020-04-05
Result list interaction

gjvera:
image


src/main/webapp/components/index-cards/index-cards.js
```diff
@@ -108,13 +108,13 @@ export const IndexCardItem = props => {
           subheader={subHeader || moment(modified).fromNow()}
           action={headerAction}
         />
-        <CardContent>
-          {metacard_owner && (
+        {metacard_owner && (
```
gjvera, line 111: Helps shrink the space between the title and the actions when there's no metacard owner.

src/main/webapp/components/workspace/workspace.js
```diff
@@ -55,16 +60,27 @@ const workspaceById = gql`
   }
 `
 
-//TODO add paging
-const Results = ({ results }) =>
+const Results = ({ results, handleOpen, setResult }) =>
```
gjvera, line 63: Should we move the result index cards to their own file? The more interactions we add the larger this will get.

src/main/webapp/components/index-cards/index-cards.js
```diff
@@ -107,14 +108,15 @@ export const IndexCardItem = props => {
           title={title}
           subheader={subHeader || moment(modified).fromNow()}
           action={headerAction}
+          titleTypographyProps={titleTypographyProps || ''}
```
gjvera, line 111: image

Currently the result index cards header makes the result index cards a little big. This will let us customize it if we ever wanted to change how the result cards look.

willwill96, line 111: :pencil2: I don't think we should have the `|| ''` because that prop is expected to be an object or undefined

src/main/webapp/components/workspace/workspace.js
willwill96: I think that's a good idea. Probably better to do in a separate pr though

willwill96: It may also be worth decoupling result index cards from list view index cards

Issue #299, opened 2020-03-26
Querying with empty advanced search results in source error


samuelechu:
![image](https://user-images.githubusercontent.com/4467636/77700783-24a57700-6f72-11ea-934f-bc8de6d5999b.png)

PR #322, opened 2020-04-08
Fix Advanced Query search error

mazarag2:
- Fix source error when doing an empty search
https://github.com/connexta/revelio/issues/299
- Add check in Results to display text if its available otherwise use an empty string


src/main/webapp/components/results/results.js
```diff
@@ -84,7 +84,7 @@ const ExpandButton = props => {
 
 const Description = props => {
   const [expanded, setExpanded] = useState(false)
-  const { text } = props
```
samuelechu, line 87: :pencil2: `const { text = '' } = props`

src/main/webapp/intrigue-api/metacards/metacards.js
```diff
@@ -151,6 +151,9 @@ const WILDCARD_FITLER = {
 }
 
 const getCql = ({ filterTree, cql }) => {
+  if (filterTree.filters && filterTree.filters.length == 0) {
```
samuelechu, line 154: This could result in an error if filterTree was undefined. Could use getIn(filterTree, ['filters','length'], 0) and check to see that it's not 0.

willwill96: We should also be using `===` just to be safe

mazarag2: The problem I was seeing is that `getCql` also gets hit when loading up the worskpaces the filterTree was in a state for `getIn(filterTree, ['filters','length'], 0) ` to be true which we dont want and  thought of an additional check if an actual filter exist rather than bieng undefined . This is whats currently working 
|  Component  | getIn(filterTree, ['filters', 'length'], 0) == 0          |  Array.isArray(getIn(filterTree, ['filters'], 0))       |
| ------------- |:-------------:|:-----:|
| Workspace load     | true | false |
| Metacard click     |  false   |  true |
| Empty Advanced Search | true |  true |


```javascript  
const checkForEmptySearch =
    Array.isArray(getIn(filterTree, ['filters'], 0)) &&
    getIn(filterTree, ['filters', 'length'], 0) == 0
```

PR #323, opened 2020-04-09
Adding result download interaction

gjvera:


src/main/webapp/components/results/result-index-cards.js
```diff
@@ -0,0 +1,47 @@
...
+  const { results, handleOpen, setResult } = props
+  return React.useMemo(
+    () =>
+      results.map(({ metacard }) => (
+        <IndexCardItem
+          key={metacard.attributes.id}
+          title={metacard.attributes.title}
+          subHeader={' '}
+          titleTypographyProps={{
+            variant: 'body1',
```
gjvera, line 17: image
Thoughts on this styling change

src/main/webapp/components/workspace/workspace.js
```diff
@@ -236,7 +210,7 @@ export default () => {
                   })
                 }}
               />
-              <Results
+              <ResultIndexCards
                 results={results}
                 handleOpen={handleOpen}
```
willwill96, line 215: :pencil2: I don't think that workspace should care about the logic related to lists.

I think we should move the ResultListInteraction and Popover components as well as the selectedResult state into the ResultIndexCards component(if not it's own).

src/main/webapp/components/results/result-index-cards.js
willwill96: lgtm - on an unrelated note, it kind of looks like the add and download buttons are slightly off center, but that could just be the way the SVGs are

src/main/webapp/components/results/result-index-cards.js
```diff
@@ -0,0 +1,91 @@
...
+import { ResultListInteraction } from '../lists/result-list-interaction'
+
+export default props => {
+  const { results, setLists, lists } = props
+  const [selectedResult, setSelectedResult] = React.useState(null)
+  const [anchorEl, handleOpen, handleClose, isOpen] = useAnchorEl()
+
+  return (
+    <React.Fragment>
+      {' '}
```
willwill96, line 17: should this be deleted?

src/main/webapp/components/results/result-index-cards.js
```diff
@@ -0,0 +1,91 @@
...
+                        metacard.attributes['resource-download-url'],
+                        '_blank'
+                      )
+                    }}
+                    size="small"
+                  >
+                    <GetAppIcon />
+                  </IconButton>
+                ) : null}
+                {props.actions ? props.actions : null}
```
willwill96, line 52: :pencil2: could also be written `props.actions || null`

src/main/webapp/components/results/result-index-cards.js
gjvera: I'm not sure why but whenever I format this comes up whenever something is wrapped in a React.Fragment. I'll remove it.

PR #329, opened 2020-04-10
Added filters to workspace results

AzGoalie:
Added the ability to filter results in a workspace.
image

### DIFF
--- a/src/main/webapp/components/workspace/filters.js
+++ b/src/main/webapp/components/workspace/filters.js
@@ -38,16 +38,23 @@ const Filters = ({ onApply, onChange }) => {
     [filters]
   )
 
-  useEffect(() => {
-    if (enable) {
-      onChange(filters)
-    } else {
-      onChange([])
-    }
-  }, [enable])
+  useEffect(
+    () => {
+      if (enable) {
+        onChange(filters)
+      } else {
+        onChange([])
+      }
+    },
+    [enable]
+  )
 
   return (
-    <Section title="Filters" enable={enable} onChange={() => setEnable(!enable)}>
+    <Section
+      title="Filters"
+      enable={enable}
+      onChange={() => setEnable(!enable)}
+    >
       <Box display="flex" alignItems="center" justifyContent="center">
         <Box padding="16px">
           <Button
--- a/src/main/webapp/components/workspace/filters.js
+++ b/src/main/webapp/components/workspace/filters.js
@@ -1,0 +1,82 @@
+import React, { useState, useEffect } from 'react'
+import Button from '@material-ui/core/Button'
+import Divider from '@material-ui/core/Divider'
+import Section from './section'
+import Typography from '@material-ui/core/Typography'
+import { defaultFilter } from '../query-builder/filter/filter-utils'
+import Box from '@material-ui/core/Box'
+import Filter from '../query-builder/filter/individual-filter'
+
+const anyTextFilter = { ...defaultFilter, value: '*' }
+
+const FilterList = ({ filters, onChange }) =>
+  filters.map((filter, i) => (
+    <Box key={i} style={{ padding: '0px 16px' }}>
+      <Filter
+        filter={filter}
+        onChange={newFilter => {
+          // Have to copy / clone array to modify
+          const newFilters = filters.slice()
+          newFilters[i] = newFilter
+          onChange(newFilters)
+        }}
+        onRemove={() => {
+          onChange(filters.filter((filter, index) => index !== i))
+        }}
+      />
+    </Box>
+  ))
+
+const Filters = ({ onApply, onChange }) => {
+  const [filters, setFilters] = useState([])
+  const [enable, setEnable] = useState(false)
+
+  useEffect(
+    () => {
+      onChange(filters)
+    },
+    [filters]
+  )
+
+  useEffect(() => {
+    if (enable) {
+      onChange(filters)
+    } else {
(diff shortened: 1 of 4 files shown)


AzGoalie:
The apply filters button just reruns the search, but im not sure if this is the best UX. I added it because it might be confusing that nothing happened when a filter was added.

Filters are applied to the search query when enabled, so it's slightly different than what Intrigue does, but @pklinef approved this version.

src/main/webapp/components/workspace/workspace.js
```diff
@@ -236,6 +250,13 @@ export default () => {
                   })
                 }}
               />
+              <Filters
+                filters={filters}
+                setFilters={setFilters}
```
samuelechu, line 255: :pencil2: onChange instead of setFilters to preserve consistency

src/main/webapp/components/workspace/workspace.js
```diff
@@ -236,6 +250,13 @@ export default () => {
                   })
                 }}
               />
+              <Filters
```
samuelechu, line 253: I think it would be better have the filters component keep track of the enabled/disabled state instead

src/main/webapp/components/workspace/workspace.js
```diff
@@ -236,6 +250,13 @@ export default () => {
...
+                filters={filters}
+                setFilters={setFilters}
+                enable={enableFilters}
+                setEnable={setEnableFilters}
+                onApply={() => onSearchQuery(currentQuery)}
```
willwill96, line 258: Should we be rerunning the query on changes to the filter? In intrigue the filtering is done client side

src/main/webapp/components/workspace/workspace.js
AzGoalie: I tried that approach first but we need to disable the filters when not enabled in the workspace component. How would you go about pushing it down?

samuelechu: In that case, you would still use the onChange to set filters in the workspace to [] , while keeping the filters stored on the Filters component.

src/main/webapp/components/workspace/workspace.js
```diff
@@ -62,10 +64,18 @@ const workspaceById = gql`
   }
 `
 
-const queryToSearch = query => {
+const queryToSearch = (query, filters) => {
   const { sources, sorts, detail_level, filterTree } = query
+
+  const searchQuery = !filterTree ? defaultQuery : filterTree
```
AzGoalie, line 70: Sometimes, when a search query is swapped with another, and the query wasn't changed from the default anyText *, it's filterTree is null

src/main/webapp/components/workspace/workspace.js
```diff
@@ -212,6 +242,12 @@ export default () => {
                     })
                   }}
                 />
+                <Filters
+                  filters={filters}
```
samuelechu, line 246: I think it would be better to just pass the onChange to the Filters component, and call onChange with the filters stored locally on that component, or [ ] when the enabled/disabled switch is toggled - should simplify `onFilterChange` and the `onSearchQuery` a bit

src/main/webapp/components/workspace/workspace.js
```diff
@@ -82,12 +85,24 @@ const queryToSearch = query => {
   }
 }
 
+const addFiltersToQuery = (query, filters) => {
+  const { filterTree } = query
+
+  const searchTree = !filterTree ? defaultQuery : filterTree
```
AzGoalie, line 91: Sometimes when changing search queries, the filter tree is null. I assume this is when a default query is created and never changed.

src/main/webapp/components/workspace/filters.js
```diff
@@ -0,0 +1,77 @@
...
+          >
+            <Typography>Add Filter</Typography>
+          </Button>
+        </Box>
+
+        <Box padding="16px">
+          <Button
+            color="primary"
+            variant="contained"
+            onClick={() => shouldAddFilters()}
```
samuelechu, line 61: I'd argue that the enabled state should be set to true whenever the apply button is pressed. You can also add in a useEffect whenever enabled is toggled that reruns the query with the correct filters. What do you think?

AzGoalie: I'm not sure it's necessary, as you won't be able to see the Apply Filters button unless enabled is set to true. If we want to make sure that filters are enabled, i can add it.

samuelechu: hmm, it doesn't look like the Apply Filters button is conditionally rendered. Where does that ^ happen?

AzGoalie: It's part of the children of Section

samuelechu: Pretty cool, didn't know about the Collapse component :+1: Shouldn't it be safe to call the onApply without checking for the enabled state then?

src/main/webapp/components/workspace/filters.js
```diff
@@ -0,0 +1,84 @@
...
+    () => {
+      onChange(filters)
+    },
+    [filters]
+  )
+
+  const toggleSection = () => {
+    if (enable) {
+      setEnable(false)
+      onChange([])
```
samuelechu, line 44: Can move the onChange to the useEffect to avoid potential issues with the setEnable being async

PR #330, opened 2020-04-10
Extract permissions logic into custom hook

samuelechu:
Was starting work on deletion of queries(which use index cards), and thought that this would be good to bring in first. Metacard interactions will also be changed to use this hook if this pr gets through.


src/main/webapp/react-hooks/use-get-permissions.tsx
```diff
@@ -0,0 +1,45 @@
...
+const user = gql`
+  query UserEmailAndRoles {
+    user {
+      email
+      roles
+    }
+  }
+`
+
+const useGetPermissions = (user: User) => {
```
willwill96, line 22: :pencil2: usePermissions

PR #331, opened 2020-04-13
Hidden/Displayed columns

gordocanchola:
Adding a Columns button to bring up the columns transferlist
The transferlist is used to moved attributes to and from Displayed and
Hidden
The Displayed list is orderable by dragging and the order results in
column order

image

### DIFF
--- a/src/main/webapp/components/results/results.js
+++ b/src/main/webapp/components/results/results.js
@@ -92,7 +92,7 @@ const TransferListModal = props => {
       .filter(column => column.hidden)
       .map(column => column.title)
 
-    if (onColumnUpdate instanceof Function) {
+    if (typeof onColumnUpdate === 'function') {
       onColumnUpdate(order, hide)
     }
 
--- a/src/main/webapp/components/results/results.js
+++ b/src/main/webapp/components/results/results.js
@@ -88,12 +88,12 @@ const TransferListModal = props => {
 
   const handleSave = () => {
     const order = columnOrder.map(column => column.title)
-    const hide = columnOrder
+    const columnHide = columnOrder
       .filter(column => column.hidden)
       .map(column => column.title)
 
     if (typeof onColumnUpdate === 'function') {
-      onColumnUpdate(order, hide)
+      onColumnUpdate({ columnOrder: order, columnHide })
     }
 
     onClose()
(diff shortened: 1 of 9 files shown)


src/main/webapp/components/results/transfer-list.js
```diff
@@ -0,0 +1,207 @@
...
+      </div>
+    </div>
+  )
+}
+
+// a little function to help us with reordering the result
+const reorder = (list, startIndex, endIndex) => {
+  const [lr] = list.splice(startIndex, 1)
+  list.splice(endIndex, 0, lr)
+  list.forEach((column, index) => (column.tableData.columnOrder = index))
```
willwill96, line 140: It's typically bad practice to modify react state arrays in place

src/main/webapp/components/results/transfer-list.js
```diff
@@ -0,0 +1,207 @@
...
+
+  return list
+}
```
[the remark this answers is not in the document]
gordocanchola: Created a list-utils and added  reorderList

src/main/webapp/components/results/transfer-list.js
gordocanchola: Changed to return a new list instead of modifying the supplied list.

src/main/webapp/components/results/results.js
```diff
@@ -278,21 +326,24 @@ const Container = props => {
     )
   }
 
+  const onColumnUpdate = (columnOrder, columnHide) => {
```
samuelechu, line 329: I think the updateUserPrefs function allows for passing in just the fields you want to update, which means you can call it with just {columnOrder, columnHide}. I noticed that the hook doesn't check to see if the userPrefs changed when the updateUserPrefs function is called. You could add that functionality into the hook if you wanted

gordocanchola: Changed to pass a map of user pref fields.

Not making changes for the hook to check if the userPrefs changed.
I don't think I have access to the currentPrefs outside of the update function which only deals with updating the cache,  not actually performing the mutation.

gordocanchola: I ended up making the changes for the hook to check for changes in userPrefs before writing.
Thanks for the suggested change, Sam.

PR #333, opened 2020-04-14
Change metacard interactions to use permissions hook

samuelechu:
Also adds context to workspaces so that security attributes can be passed down for query cards to show available actions. https://github.com/connexta/revelio/pull/333/commits/904b35a44d5821b60a3ac11e81e542bda775d188

### DIFF
--- a/src/main/webapp/components/result-forms/route.js
+++ b/src/main/webapp/components/result-forms/route.js
@@ -1,36 +1,26 @@
-import React, { Fragment, useState } from 'react'
-
-import gql from 'graphql-tag'
-import { useQuery, useMutation } from '@apollo/react-hooks'
-
-import ResultForms from './result-forms'
-
+import { useMutation, useQuery } from '@apollo/react-hooks'
 import Dialog from '@material-ui/core/Dialog'
-
 import LinearProgress from '@material-ui/core/LinearProgress'
-import { Notification } from '../notification/notification'
+import gql from 'graphql-tag'
+import React, { Fragment, useState } from 'react'
 import {
-  getSecurityAttributesFromMetacard,
-  getPermissions,
-} from '../sharing/sharing-utils'
-
-import {
-  IndexCards,
-  IndexCardItem,
+  Actions,
   AddCardItem,
   DeleteAction,
-  ShareAction,
-  Actions,
+  IndexCardItem,
+  IndexCards,
   ReadOnly,
+  ShareAction,
 } from '../index-cards'
-import RetryComponent from '../network-retry/snackbar-retry'
-
 import {
-  MetacardInteractionsDropdown,
-  ShareMetacardInteraction,
   ConfirmDeleteMetacardInteraction,
   EditMetacardInteraction,
+  MetacardInteractionsDropdown,
+  ShareMetacardInteraction,
 } from '../index-cards/metacard-interactions'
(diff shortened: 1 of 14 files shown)


src/main/webapp/components/result-forms/route.js
```diff
@@ -91,7 +73,7 @@ const Item = props => {
         </Dialog>
       ) : null}
       <IndexCardItem {...form} onClick={() => setEditing(true)}>
-        <Actions attributes={form} user={user}>
+        <Actions attributes={form}>
```
samuelechu, line 76: Going to see if I can use context in a future pr which will allow a user to be passed through to the hook in storybook

src/main/webapp/components/workspace/workspace.js
```diff
@@ -138,121 +144,123 @@ export default () => {
...
   const { title } = attributes
   return (
-    <div
-      style={{
-        display: 'flex',
-        height: 'calc(100vh - 64px)',
-        overflow: 'hidden',
-      }}
-    >
+    <WorkspaceContext.Provider value={attributes}>
```
samuelechu, line 147: The only thing that changed below this line is the closing tag of WorkspaceContext.Provider

src/main/webapp/components/search-forms/search-forms.stories.tsx
```diff
@@ -44,10 +44,10 @@ const startingForms = [
...
 ]
 
-const userAttributes = {
-  email: 'admin@test.com',
-  roles: ['admin'],
-}
+// const userAttributes = {
+//   email: 'admin@test.com',
+//   roles: ['admin'],
+// }
```
willwill96, line 50: delete?

src/main/webapp/react-hooks/use-get-permissions.tsx
```diff
@@ -19,12 +26,12 @@ const user = gql`
...
-const usePermissions = (user: User) => {
-  return (attributes: any) => {
+const usePermissions = (user: User = {}) => {
+  return (attributes: any = {}) => {
     const securityAttributes = getSecurityAttributesFromMetacard(attributes)
     return getPermissions(
-      user.email,
-      user.roles,
+      user.email || '',
+      user.roles || [],
```
willwill96, line 34: with these changes do you still need, this change? https://github.com/connexta/revelio/pull/333/files#diff-8e61c7dfd17cef5f94afbba276fdbc9eR56

samuelechu: An undefined value still may be returned if the user query is still loading https://github.com/connexta/revelio/pull/333/files/e8340ec91a66c723502d36849e1ebd797cb8eabd#diff-fc3b9df5837400effc9fe86ea4d16242R44

src/main/webapp/components/search-forms/search-forms.stories.tsx
samuelechu: I'm hoping to uncomment this later, once I have the usePermissions hook able to take in a user through context

samuelechu:
> I think user still needs to be removed from the workspaces.gql file as well as the workspaces use-create hook

This is still using user. Do you think it would be good to create a hook for just user?https://github.com/connexta/revelio/blob/master/src/main/webapp/components/workspaces/workspaces.js#L164

PR #337, opened 2020-04-17
Bulk export

gjvera:


src/main/webapp/components/results/result-index-cards.js
```diff
@@ -8,12 +8,34 @@ import useAnchorEl from '../../react-hooks/use-anchor-el'
...
+  } = props
   const [selectedResult, setSelectedResult] = React.useState(null)
   const [anchorEl, handleOpen, handleClose, isOpen] = useAnchorEl()
 
+  const modifySelectedResults = result => {
+    const index = resultsToExport.indexOf(
+      item => result.attributes.id === item.attributes.id
+    )
+    if (index === -1) {
+      const updatedResults = resultsToExport
```
gjvera, line 29: I tried doing 
`setResultsToExport([...resultsToExport, result])` here but the sate was never getting updated. That's what I did it this way. Not sure if I was doing something wrong.

src/main/webapp/components/result-export/result-export.js
```diff
@@ -114,9 +112,53 @@ const ResultExport = props => {
       <Button
         disabled={selectedFormat === '' ? true : false}
         fullWidth
-        onClick={() => {
+        onClick={async () => {
           const encodedTransformer = getExportFormatId(selectedFormat)
-          props.saveFile({ result, exportResult, encodedTransformer })
+          const cql = getResultSetCql(resultsToExport)
```
willwill96, line 117: frontend shouldn't have knowledge of cql. We should be passing in a filter tree and having graphql handle the conversion

src/main/webapp/components/result-export/result-export.js
```diff
@@ -114,9 +112,53 @@ const ResultExport = props => {
...
+            res = await exportResultSet({
+              variables: {
+                transformer: 'zipCompression',
+                body: {
+                  cql,
+                  srcs,
+                  count,
+                  args: {
+                    transformerId: encodedTransformer,
+                  },
```
willwill96, line 131: I think we should type this object out in graphql and not use the generic Json

src/main/webapp/components/workspace/workspace.js
```diff
@@ -89,6 +89,7 @@ export default () => {
   const [currentQuery, setCurrentQuery] = useState(null)
   const [lists, setLists] = React.useState(null)
   const [queries, setQueries] = useState()
+  const [resultsToExport, setResultsToExport] = React.useState([])
```
willwill96, line 92: I think we should consider having this state managed by context instead of passing it down through props. Similar to how the [selection interface](https://github.com/connexta/revelio/blob/master/src/main/webapp/react-hooks/use-selection-interface.js) works. @samuelechu thoughts?

src/main/webapp/components/result-export/result-export.js
```diff
@@ -27,33 +29,23 @@ const useExportMutation = () => {
...
-  const { loading, error, data, refetch } = useQuery(getExportOptions, {
-    variables: { transformerType: 'metacard' },
+  const transformerType =
+    !props.zipped && props.resultsToExport.length > 1 ? 'query' : 'metacard'
+  const { loading, error, data, refetch } = useQuery(exportOptions, {
+    variables: { transformerType },
   })
-
   const [exportResultHook] = useExportMutation()
+  const [exportResultSetHook] = useExportSetMutation()
```
willwill96, line 48: :pencil2: 
```suggestion
  const [exportResult] = useExportMutation()
  const [exportResultSet] = useExportSetMutation()
```

src/main/webapp/components/results/result-index-cards.js
willwill96: I think this is because of the React.memo call on results. The list of results won't re-render unless results change: https://github.com/connexta/revelio/blob/master/src/main/webapp/components/results/result-index-cards.js#L59

When you update the state to a new array, results are the same, the list of results doesn't re-render, and so the individual results still have access to the old array. That's why pushing to that array works (but we shouldn't be doing it that way). One solution would be to add resultsToExport to the react memo dependency array so that it re-renders on changes to that array.

That being said, if you go the route of using the selection interface internally in the exportaction and checkbox, you won't need to worry about it because they will re-render independently of the result list (I think this is the best option).

src/main/webapp/components/query-manager/query-manager.tsx
```diff
@@ -51,6 +56,15 @@ const QueryCard = (props: QueryCardProps) => {
         <Actions attributes={attributes}>
           <EditAction onEdit={handleOpen} />
           <DeleteAction />
+          <MetacardInteractionsDropdown>
+            <CompressedExportMetacardInteraction
+              resultsToExport={props.resultsToExport}
+              zipped={true}
```
willwill96, line 62: :pencil2: looks like this prop is unused

willwill96, line 62: :pencil2: I think this prop should be moved directly into CompressedExportInteraction because the name implies it will be zipped

src/main/webapp/components/results/result-index-cards.js
gjvera: After investigating this for a while, I've decided to break that out into another pr. Since the export calls to ddf require the list of sources to each result, I would need to add the sourceId the selectioninterface as opposed to just the ids. This would require touching all the components that currently use the selectionInterface.

src/main/webapp/components/workspace/workspace.js
gjvera: See above comment

src/main/webapp/components/query-manager/query-manager.tsx
```diff
@@ -21,6 +22,15 @@ const {
   ExportMetacardInteraction,
   CompressedExportMetacardInteraction,
 } = require('../result-export/result-export-action')
+import gql from 'graphql-tag'
+
+const getSources = gql`
+  query queryExecutorSources {
```
willwill96, line 28: :pencil2: rename to avoid collisions

src/main/webapp/components/results/result-index-cards.js
```diff
@@ -53,12 +35,16 @@ export default props => {
               <Actions disableSpacing={true}>
                 <Checkbox
```
willwill96, line 36: I would recommend making a separate custom checkbox component that uses the selection interface hook internally - that way you don't need to add selection to the React.memo dependency array

src/main/webapp/components/result-export/result-export.js
```diff
@@ -49,8 +50,12 @@ const useExportSetMutation = () => {
 }
 
 const Container = props => {
+  const [selection, onSelect] = useSelectionInterface()
+  const resultIds = [...selection]
```
willwill96: :pencil2: would this work as well?
```suggestion
  const [resultIds, onSelect] = useSelectionInterface()
```

gjvera: Since the selection interface is a set rather than an array and the backend expects an array it has to be converted  with the `...` operator

PR #339, opened 2020-04-18
Editable workspace title

gjvera:
image

image

image

### DIFF
--- a/src/main/webapp/components/workspace/hooks/use-save-workspace.tsx
+++ b/src/main/webapp/components/workspace/hooks/use-save-workspace.tsx
@@ -22,7 +22,7 @@ export default () => {
     const queries = (workspace.queries || []).map((query: QueryType) => ({
       id: query.id,
     }))
-    const res = save({
+    save({
       variables: {
         id,
         attrs: {
@@ -33,6 +33,5 @@ export default () => {
       },
     })
     setSaving(false)
-    console.log(res)
   }
 }
--- a/src/main/webapp/components/workspace/hooks/use-save-workspace.tsx
+++ b/src/main/webapp/components/workspace/hooks/use-save-workspace.tsx
@@ -17,11 +17,12 @@ export default () => {
     }
   `
   const [save] = useMutation(mutation)
-  return (workspace: Workspace) => {
+  return (setSaving: (isSaving: boolean) => void, workspace: Workspace) => {
+    setSaving(true)
     const queries = (workspace.queries || []).map((query: QueryType) => ({
       id: query.id,
     }))
-    save({
+    const res = save({
       variables: {
         id,
         attrs: {
@@ -31,5 +32,7 @@ export default () => {
         },
       },
     })
+    setSaving(false)
+    console.log(res)
   }
 }
--- a/src/main/webapp/components/workspace/hooks/use-save-workspace.tsx
+++ b/src/main/webapp/components/workspace/hooks/use-save-workspace.tsx
@@ -17,11 +17,7 @@ export default () => {
     }
   `
   const [save] = useMutation(mutation)
-  return async (
-    setSaving: (isSaving: boolean) => void,
-    workspace: Workspace
-  ) => {
-    setSaving(true)
+  return async (workspace: Workspace) => {
     const queries = (workspace.queries || []).map((query: QueryType) => ({
       id: query.id,
     }))
@@ -35,6 +31,5 @@ export default () => {
         },
       },
     })
-    setSaving(false)
   }
 }
--- a/src/main/webapp/components/workspace/hooks/use-save-workspace.tsx
+++ b/src/main/webapp/components/workspace/hooks/use-save-workspace.tsx
@@ -17,12 +17,15 @@ export default () => {
     }
   `
   const [save] = useMutation(mutation)
-  return (setSaving: (isSaving: boolean) => void, workspace: Workspace) => {
+  return async (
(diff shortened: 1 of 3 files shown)


Schachte:
👍

src/main/webapp/components/workspace/hooks/use-save-workspace.tsx
```diff
@@ -17,11 +17,15 @@ export default () => {
     }
   `
   const [save] = useMutation(mutation)
-  return (workspace: Workspace) => {
+  return async (
```
willwill96, line 20: While I agree this hook should be made async, so that you can manage state of requests, I don't think that we should be passing in a function to update the loading state.  I think the workspace should be handling the state on it's own.

src/main/webapp/components/workspace/workspace.js
```diff
@@ -160,9 +172,50 @@ export default () => {
...
+            style={{
+              display: 'flex',
+              alignItems: 'baseline',
+            }}
+          >
+            <TextField
+              variant="outlined"
+              style={{ margin: '10px', width: '70%' }}
+              value={title}
+              onChange={e => setTitle(e.target.value)}
```
willwill96, line 185: I don't think we should introduce state for title. I would suggest using an uncontrolled input here by using a `defaultValue` prop of `attributes.title` and having the onBlur reference the target value. i.e. 
```
onBlur={e=>saveWorkspace(..., {title: e.target.value} ...)}
```

src/main/webapp/components/workspace/hooks/use-save-workspace.tsx
gjvera: I was considering both ways. I decided to do it this way so the saving icon is always shown whenever the hook is called. That way devs in the future don't have to worry about setting the saving state.

src/main/webapp/components/workspace/workspace.js
gjvera: done 👍

src/main/webapp/components/workspace/hooks/use-save-workspace.tsx
samuelechu: I think what @willwill96 suggested is the better option. You don't need to worry about setting the saving state if you're just passing it back through the hook. You can return [saveWorkspace, saving] from the hook, and the component that needs the saving state could render based on the saving variable.

src/main/webapp/components/workspace/workspace.js
```diff
@@ -160,9 +171,49 @@ export default () => {
...
+                <CheckCircleIcon
+                  style={{ marginTop: '10px' }}
+                  color="primary"
+                />
+                <Typography variant="body1" style={{ paddingLeft: 10 }}>
+                  Saved
+                </Typography>
+              </div>
+            )}
+          </div>
```
samuelechu, line 216: I think it would be good to pull this out into its own component

src/main/webapp/components/workspace/workspace.js
```diff
@@ -160,9 +171,49 @@ export default () => {
...
+              display: 'flex',
+              alignItems: 'baseline',
+            }}
+          >
+            <TextField
+              variant="outlined"
+              style={{ margin: '10px', width: '70%' }}
+              defaultValue={title}
+              onBlur={e => {
+                saveWorkspace(setSaving, { title: e.target.value, queries })
```
samuelechu, line 185: Might be good to have a function that returns the current workspace merged with updated properties you pass the function, instead of getting all the need properties manually

willwill96: IIRC the graphql resolver supports partial updates. I think the only reason you need to pass in queries, is because it defaults to an empty array in the hook: https://github.com/connexta/revelio/blob/master/src/main/webapp/components/workspace/hooks/use-save-workspace.tsx#L21-L23

samuelechu: I tried doing the saveMetacard mutation above with just the title, and it removed all the queries on the workspace :crying_cat_face:

gjvera: @samuelechu the reason that is, is because `use-save-workspace` has the following code
```const queries = (workspace.queries || []).map((query: QueryType) => ({
      id: query.id,
}))```
so if no queries are passed in it defaults to an empty array. Hence the removal of all queries in the workspace if you pass just the title.

samuelechu: Here's the query I did in graphiql (`saveQueryToWorkspace`). 
image

Would you verify that you're able to do a partial update? It didn't happen for me.

samuelechu: if I removed queries from `workspaceAttrs`, all queries were removed from the workspace. When I removed title however, the current title was preserved.

samuelechu: > @samuelechu the reason that is, is because `use-save-workspace` has the following code
> 
> ```
> const queries = (workspace.queries || []).map((query: QueryType) => ({
>       id: query.id,
> }))
> ```
> 
> so if no queries are passed in it defaults to an empty array. Hence the removal of all queries in the workspace if you pass just the title.

My mistake, I didn't clarify that I was doing a mutation in graphiql

willwill96: I think this is a bug with our graphql, when it sends the save request, queries is a function: 
image

PR #349, opened 2020-04-23
Adding result interaction dropdown

gjvera:
image


src/main/webapp/components/metacard-interaction/metacard-interactions-dropdown.tsx
```diff
@@ -20,6 +20,7 @@ const MetacardInteractionsDialogContext = createContext<any>({
 
 type MetacardInteractionsDropdownProps = {
   children: React.ReactNode
+  size?: 'small' | 'medium'
```
willwill96, line 23: probably fine for now, but to be able to add more props to the icon we may want to generalize this prop to an object called `iconProps` of type [IconProps](https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Icon/Icon.d.ts)

Issue #345, opened 2020-04-22
Selecting a source to search on for simple-search queries all sources


mazarag2:

mazarag2:
![search](https://user-images.githubusercontent.com/16808210/80014129-b9719680-8484-11ea-9435-38d9b93d4fb2.png)

PR #352, opened 2020-04-23
Fix Query in simple search when selecting sources

mazarag2:
When doing a simple search for a specific source all sources are still queried 
Fixes: #345

When use-query-executor was bieng called for `onSearch` a check exists wether or not `sourceIds` exist or not. if they dont query all sources. in Simple Search `sources` was being passed rather than `sourceIds` which caused it to query all sources

### DIFF
--- a/src/main/webapp/components/basic-search/basic-search.js
+++ b/src/main/webapp/components/basic-search/basic-search.js
@@ -266,7 +266,7 @@ const createQuery = filterMap => {
 
   return {
     filterTree: toFilterTree(filterMap),
-    sources: sources || null,
+    sourceIds: sources || null,
     sorts,
   }
 }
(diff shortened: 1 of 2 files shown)


src/main/webapp/components/basic-search/basic-search.js
```diff
@@ -266,7 +266,7 @@ const createQuery = filterMap => {
 
   return {
     filterTree: toFilterTree(filterMap),
-    sources: sources || null,
+    sourceIds: sources || null,
```
samuelechu, line 269: This might break searches using the basic search form in workspaces https://github.com/connexta/revelio/blob/master/src/main/webapp/components/workspace/workspace.js#L65. Could you instead transform the sources property for the query in simple search?

src/main/webapp/components/simple-search/simple-search.js
```diff
@@ -69,6 +75,7 @@ const SimpleSearch = props => {
               setPageIndex(0)
               setQuery(query)
               onClear()
+              transformToSourceIds(query)
```
samuelechu, line 78: :pencil2: This function could return a new object that has the desired structure. I think it's generally better not to mutate things if you can help it

PR #353, opened 2020-04-23
Adds deletion of queries from workspace

samuelechu:


src/main/webapp/components/query-manager/query-manager.tsx
```diff
@@ -56,7 +56,7 @@ const QueryCard = (props: QueryCardProps) => {
       >
         <Actions attributes={attributes}>
           <EditAction onEdit={handleOpen} />
-          <DeleteAction />
+          <DeleteAction itemName="query" onDelete={onDelete} />
```
gjvera, line 59: Should we put this in the `MetacardInteractionsDropdown` similar to how DDF does it? When we add the other icons, it may look a little crowded in the actions area.

samuelechu: We can move it when it gets too crowded. Do we know if we definitely want the icons to be exactly the same as they are in intrigue? It seems that icons like the play button wouldn't be as useful on there.

gjvera: Probably warrants a discussion with the ux team.

src/main/webapp/components/workspace/workspace.js
```diff
@@ -91,6 +97,17 @@ export default () => {
 
   const { results, status, onSearch, onCancel, onClear } = useQueryExecutor()
 
+  const deleteQuery = useDeleteQuery(id => {
+    const updatedQueries = queries.filter(query => query.id != id)
```
willwill96, line 101: :pencil2: 
```suggestion
    const updatedQueries = queries.filter(query => query.id !== id)
```

Issue #357, opened 2020-04-24
Make table search more clear


gjvera:
Currently the table search just says "Search." This could confuse some users. Make it more clear by labeling it "Search Results" or "Filter Results."

PR #359, opened 2020-04-26
Addressing some ux feedback

### LINKED ISSUE #354: Add arrow or icon next to "Filter Workspaces"
image

Currently this looks just like a label, add an icon next to it to let the user know it's clickable.

gjvera:
Fixes: #357 
image

Fixes: #358 
image

Fixes #354 
image


src/main/webapp/components/workspaces/filter-workspaces.js
```diff
@@ -49,12 +51,15 @@ const FilterWorkspaces = ({ onFilter, userAttributes }) => {
 
   return (
     <React.Fragment>
-      <List component="nav">
+      <List component="nav" style={{ maxWidth: '14%' }}>
```
willwill96, line 54: :pencil2: this seems like an arbitrary number - wonder if we could do something like `width: fit-content`

PR #361, opened 2020-04-27
Create search button should always be visible in a workspace

samuelechu:
image
The create search button was not showing when the 'Lists' tab was active


src/main/webapp/components/query-manager/add-query.tsx
```diff
@@ -36,7 +35,7 @@ export default (props: AddQueryProps) => {
     <React.Fragment>
       <CreateSearch />
 
-      {!hasQueries && (
+      {showSecondarySection && (
```
willwill96, line 38: This prop feels a little awkward. Could we pull this div out of this component or pass it in as render props?

PR #365, opened 2020-04-30
Adds refresh result count interaction

samuelechu:
image

### DIFF
--- a/src/main/webapp/components/workspace/workspace.js
+++ b/src/main/webapp/components/workspace/workspace.js
@@ -165,13 +165,13 @@ export default () => {
     },
   })
 
-  const onSearchQuery = id => {
+  const onSearchQuery = (id, resultCountOnly) => {
     const query = queries.find(query => query.id === id)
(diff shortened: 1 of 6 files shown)


willwill96:
in intrigue, i think this refreshes the result count without running the search. not sure if we want to maintain this behavior

bdeining:
> in intrigue, i think this refreshes the result count without running the search. not sure if we want to maintain this behavior

to my knowledge it runs the search to get the count, but does not parse the results.   some sources are equipped to do a `hit count only` query.

src/main/webapp/components/workspace/workspace.js
```diff
@@ -87,6 +91,7 @@ const queryToSearch = query => {
         sortOrder: sort.substring(splitIndex + 1, sort.length),
       }
     }),
+    ...pagingFields,
```
samuelechu, line 94: It appears that a search that sets pageSize to zero runs faster than one that only returns the status, not sure why

PR #371, opened 2020-05-13
Attribute Aliases

gordocanchola:
Adding attribute aliases to the results table.
Using the attributeAliases map from systemProperties.
Updating the results table story with attribute aliases

### DIFF
--- a/src/main/webapp/components/results/results.js
+++ b/src/main/webapp/components/results/results.js
@@ -1,5 +1,7 @@
 import React, { useState, memo } from 'react'
-import { Set, fromJS, OrderedSet } from 'immutable'
+import { Set, fromJS, OrderedSet, getIn } from 'immutable'
+import gql from 'graphql-tag'
+import { useQuery } from '@apollo/react-hooks'
 import {
   useKeyPressed,
   useSelectionInterface,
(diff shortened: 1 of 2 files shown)


src/main/webapp/components/results/results.js
```diff
@@ -322,7 +347,9 @@ const Container = props => {
     { error, loading, refetch },
   ] = useUserPrefs()
 
-  if (loading) {
+  const aliasResults = useQuery(query)
```
gjvera, line 350: ✏️ Could also do `const { data: aliasData, loading: aliasLoading, error: aliasError } = useQuery(query)` or something like that to avoid doing `aliasResults.loading` `aliasResults.error` and `aliasResults.data`

src/main/webapp/components/results/results.js
```diff
@@ -313,6 +330,14 @@ const Results = props => {
   )
 }
 
+const query = gql`
```
gjvera, line 333: Since this file is kind of large, can we give a more descriptive name to this query?

gordocanchola: renamed to attributeAliasQuery

src/main/webapp/components/results/results.js
gordocanchola: Changed to rename on destructure

src/main/webapp/components/results/results.js
```diff
@@ -322,7 +354,14 @@ const Container = props => {
     { error, loading, refetch },
```
willwill96, line 354: :pencil2: might want to rename these to be prepended with userPrefs since there are multiple loading states now

gordocanchola: Renamed with userPrefs prepended

